PR #3, opened 2023-09-03
Creates Initial OAPI Spec for Pacta Versions

gbdubs:
This is an intentional small PR that:

1 - deletes the old root `pacta` package
2 - creates a basic OAPI spec for pacta versions + users
3 - does some light renaming for clarity

I want to make sure this API spec looks reasonable before going forward.


openapi/user.yaml
bcspragu, line 1: You're going to want to leave this file intact, otherwise the frontend won't be able to regenerate (e.g. `npm run generate:openapi`) bindings for the user service

bcspragu (approved):

openapi/user.yaml
gbdubs: Ahh that makes sense. I was having trouble connecting those together. THanks!

cmd/server/main.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/RMI/pacta/cmd/server/pactasrv"
 	"github.com/RMI/pacta/db/sqldb"
 	"github.com/RMI/pacta/keyutil"
-	"github.com/RMI/pacta/openapi/pacta"
+	oapipacta "github.com/RMI/pacta/openapi/pacta"
```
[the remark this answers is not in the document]
gbdubs: I initially got those confused, which is why I like the rename here. If we're going to have a generic package name, it should probably be used consistently for the same thing.

PR #5, opened 2023-09-05
Completes Initial DB Layer

gbdubs:


db/sqldb/cursor.go
```diff
@@ -7,8 +7,8 @@ import (
 	"github.com/RMI/pacta/db"
 )
 
-func offsetToCursor(i int) string {
-	return fmt.Sprintf("%d", i)
+func offsetToCursor(i int) db.Cursor {
+	return db.Cursor(fmt.Sprintf("%d", i))
```
bcspragu, line 11: Sorry, didn't notice this before:

```suggestion
	return db.Cursor(strconv.Itoa(i))
```

It's also 5x faster...not that performance matters here, but think of all the μL CO2 emissions we're preventing

db/sqldb/incomplete_upload.go
```diff
@@ -0,0 +1,223 @@
...
+		err = d.exec(tx, `DELETE FROM incomplete_upload WHERE id = $1;`, id)
+		if err != nil {
+			return fmt.Errorf("deleting incomplete_upload: %w", err)
+		}
+		buri, err = d.DeleteBlob(tx, iu.Blob.ID)
+		if err != nil {
+			return fmt.Errorf("deleting blob: %w", err)
+		}
+		return nil
+	})
```
bcspragu, line 124: This could also be a single `DELETE FROM incomplete_upload WHERE id = $1 RETURNING blob_id` instead of three separate database calls, though I care less about this one than the one deleting N child rows.

db/sqldb/migrations/0003_domain_types.up.sql
bcspragu, line 1: [no action required, this is pure snark]

Uh oh, already modifying existing migrations I see

On a slightly more serious note though, this _is_ one of the problems of front-loading the whole DB layer - it's harder to do a thorough review because it's just hundreds of lines of nearly identical SQL statements, not in the context of some PR like "Implement portfolio group members" that make omissions like this more obvious. Basically go/smallcls

db/sqldb/owner.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (d *DB) GetOrCreateOwnerForInitiative(tx db.Tx, iID pacta.InitiativeID) (pacta.OwnerID, error) {
+	owner, err := d.ownerByInitiative(tx, iID)
+	if err == nil {
+		return owner.ID, nil
+	}
+	if !db.IsNotFound(err) {
+		return "", fmt.Errorf("querying owner by user: %w", err)
+	}
+	return d.createOwner(tx, &pacta.Owner{Initiative: &pacta.Initiative{ID: iID}})
```
bcspragu, line 92: This feels like a DB endpoint that would benefit from a `StartOrContinueTransaction` block or whatever we call it now

db/sqldb/populate.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/RMI/pacta/db"
+	"github.com/RMI/pacta/pacta"
+)
+
+type HasBlobs interface {
+	Blobs() []*pacta.Blob
+}
+
+func (d *DB) PopulateBlobs(tx db.Tx, hbs []HasBlobs) error {
```
bcspragu, line 14: I don't think the database layer is the right location for this type of pointer twiddling, the telltale sign is that it doesn't actually do anything specific to the Postgres DB implementation. It could comfortably live at the server layer.

Also, do we need this whole paradigm here? I can understand it in a GraphQL server where you have arbitrarily nested things, but in a more REST-like server, you always know exactly what data you're returning and can just populate it in one fell swoop, there's no dynamicness

deps.bzl
bcspragu, line 1: Oh gross, all the GraphQL dependencies got pulled in. That makes sense, cuz `gqlerr`, but yeah we don't want these. I'll remove them when I swap out `gqlerr` for something else.

bcspragu (approved):
Question for this whole PR: Do we know we'll actually need _all_ of these database endpoints in one server endpoint or another? It's a big API surface to maintain, refactor, test, etc, esp if we aren't using it all.

cmd/server/pactasrv/pacta_version.go
```diff
@@ -31,7 +32,7 @@ func (s *Server) CreatePactaVersion(ctx context.Context, request api.CreatePacta
 		Digest:      request.Body.Digest,
 	})
 	if err != nil {
-		return nil, zap.Error(ctx, "failed to create PACTA version", zap.Error(err))
+		return nil, gqlerr.Internal(ctx, "failed to create PACTA version", zap.Error(err))
```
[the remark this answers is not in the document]
gbdubs: SG - removed. In #14 I implement something that sounds pretty close to what we want here - PTAL there.

db/sqldb/analysis_artifact_test.go
```diff
@@ -0,0 +1,124 @@
...
+	p1 := portfolioForTestingWithKey(t, tdb, "Portfolio1")
+	p2 := portfolioForTestingWithKey(t, tdb, "Portfolio2")
+	pg := portfolioGroupForTesting(t, tdb, o)
+	portfolioGroupMembershipForTesting(t, tdb, pg, p1)
+	portfolioGroupMembershipForTesting(t, tdb, pg, p2)
+	s := snapshotPortfolioGroupForTesting(t, tdb, pg)
+	b1 := blobForTestingWithKey(t, tdb, "blob1")
+	b2 := blobForTestingWithKey(t, tdb, "blob2")
+	b3 := blobForTestingWithKey(t, tdb, "blob3")
+	aid, err := tdb.CreateAnalysis(tx, &pacta.Analysis{
```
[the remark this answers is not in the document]
gbdubs: lol agreed. I'll fully standardize to that in a followup PR.

db/sqldb/cursor.go
gbdubs: Good catch.

db/sqldb/incomplete_upload.go
gbdubs: Yeah good thought - here I think the argument cuts the other way (see above) so I'll skip.

db/sqldb/migrations/0003_domain_types.up.sql
gbdubs: I hear your opinion here, I think we just have different ones. The mental overhead of context switching for me is large (perhaps larger than for you) so implementing a layer at a time has a really considerable speed up for my capacity to build out features when they're reasonably well understood (as we have in this case). I'm sure this isn't the last small drift that I'll pay for, but thus far, it feels to me like it will be a worthwhile tradeoff.

db/sqldb/owner.go
gbdubs: Good call.

db/sqldb/populate.go
gbdubs: Very good point - I'll remove this.

deps.bzl
gbdubs: Thanks for pointing it out - I've fixed this in this PR, removed these deps. Check out the err handling in #14  - I think it's more along the lines of what you want?

db/sqldb/analysis.go
```diff
@@ -0,0 +1,237 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("performing analysis sub-entity deletion: %w", err)
+	}
+	return buris, nil
+}
+
+func rowToAnalysis(row rowScanner) (*pacta.Analysis, error) {
+	a := &pacta.Analysis{
+		Owner:             &pacta.Owner{},
```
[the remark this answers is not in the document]
gbdubs: Yeah happy to discuss this. I find it most useful at the bizlogic layer, where we can do in-place data population and data conversion without creating new structs.

RE: what sets of fields are populated, it's always easy: we ONLY ever populate the ID of a nested object. Easy to maintain!

db/sqldb/analysis.go
```diff
@@ -0,0 +1,237 @@
...
+		if err != nil {
+			return fmt.Errorf("reading analysis artifacts: %w", err)
+		}
+		for _, aa := range aas {
+			buri, err := d.DeleteAnalysisArtifact(tx, aa.ID)
+			if err != nil {
+				return fmt.Errorf("deleting analysis artifact %q: %w", aa.ID, err)
+			}
+			buris = append(buris, buri)
+		}
```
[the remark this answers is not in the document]
gbdubs: Ah - now I remember why. It's because `blob_uri` is a feature on `blob` not on `artifact`. This does the trick though.

```
                 rows, err := d.query(tx, `
			WITH deleted_analysis_artifacts AS (
				DELETE FROM analysis_artifact
				WHERE analysis_id = your_analysis_id_here
				RETURNING blob_id
			)
			DELETE FROM blob
			WHERE id IN (SELECT blob_id FROM deleted_analysis_artifacts)
			RETURNING blob_uri;
		`, id)
		if err != nil {
			return fmt.Errorf("reading analysis_artifacts: %w", err)
		}
```

db/sqldb/analysis.go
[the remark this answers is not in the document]
bcspragu: Personally, I think the "in-place data population" ends up being more trouble than its worth, with all the pointer twiddling finicky-ness and what not. As for data conversion, I think it's totally reasonable for our domain types and our wire responses to have different (i.e. not nested vs nested) formats, and then doing something like `conv.AnalysisToWire(analysis, owner, ...)`. Just prevents us from going "oh cool I can just do `analysis.Owner.Name`" when that field will quite literally never be set (until we're shlepping it off to the wire format)

PR #7, opened 2023-09-07
Frontend Scaffolding + Helpers

gbdubs:
Aside from the one issue with OAPI spec discussing offline, this PR is ready for reveiw. What does it include?

- Utitilites, some standard, some new, for common frontend work.
- Some global imports (computed, onMounted, presentX)
- Helpful common display utils for forms

And, an initial editor for PACTA version definitions, including edit, delete, etc. Will need a bit more work after the OAPI issues are resolved.

Feel free to review!


frontend/pages/admin/pacta-version/index.vue
```diff
@@ -0,0 +1,56 @@
...
+  `${prefix}.deletePactaVersion`
+)
+const newPV = () => router.push('/admin/pacta-version/new')
+</script>
+
+<template>
+  <StandardContent>
+    <TitleBar title="PACTA Versions" />
+    <p>
+      General ideas about pacta versions go here.
```
bcspragu, line 24:
```suggestion
      General ideas about PACTA versions go here.
```

frontend/pages/admin/pacta-version/index.vue
```diff
@@ -0,0 +1,56 @@
...
+const router = useRouter()
+const { pactaClient } = useAPI()
+const { error: { withLoadingAndErrorHandling, handleOAPIError } } = useModal()
+
+const prefix = 'admin/pacta-version'
+const pactaVersions = useState<PactaVersion[]>(`${prefix}.pactaVersions`, () => [])
+
+const deletePV = (id: string) => withLoadingAndErrorHandling(
+  () => pactaClient.deletePactaVersion(id)
+    .then(handleOAPIError)
```
bcspragu, line 13: This is a great utility, though I think with a bit of JavaScript trickery, you could create a new type that exposes all the same functions as the `pactaClient`, but calls handleOAPIError on the result. Alternatively, we can reconfigure our API definitions or TS codegen to throw an error on non-200s and catch those as is.

If I'm understanding correctly, we'll probably want `handleOAPIError` on every single `pactaClient.whatever` call (unless something wants to handle the error explicitly), so I'd like to have something more bulletproof, as we'll invariably forget it.

frontend/pages/admin/pacta-version/new.vue
```diff
@@ -0,0 +1,53 @@
...
+const discard = () => router.push('/admin/pacta-version')
+const save = () => withLoadingAndErrorHandling(
+  () => pactaClient.createPactaVersion(pactaVersion.value).then(handleOAPIError),
+  `${prefix}.save`
+)
+</script>
+
+<template>
+  <StandardContent>
+    <TitleBar title="New PACTA Version" />
```
bcspragu, line 27: Same note here as in `StandardContent`, every page shouldn't have to specify these components, we can hoist them into the default layout and set the title with a `useTitle` composable or similar.

bcspragu (approved):

frontend/pages/admin/pacta-version/new.vue
gbdubs: I disagree on this one - mostly because <TitleBar> is likely to contain other stuff (like a 3-dot menu) that benefits from being set up as a normal component (i.e. access to computed values in it's props)

PR #14, opened 2023-09-11
PACTA Version Management (CRUD) + More Frontend Infrastructure

gbdubs:


cmd/server/pactasrv/conv_oapi_to_pacta.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"regexp"
+
+	api "github.com/RMI/pacta/openapi/pacta"
+	"github.com/RMI/pacta/pacta"
+)
+
+var initiativeIDRegex = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
+func initiativeCreateToPACTA(i *api.InitiativeCreate) (*pacta.Initiative, error) {
```
bcspragu, line 13: I think we've got enough overload on `PACTA` (the server, domain package, versions, etc), maybe we call the functions `fromOAPI` and `toOAPI` to indicate direction less ambiguously

cmd/server/pactasrv/initiative.go
```diff
@@ -0,0 +1,164 @@
...
+func (s *Server) CreateInitiative(ctx context.Context, request api.CreateInitiativeRequestObject) (api.CreateInitiativeResponseObject, error) {
+	err := s.createInitiative(ctx, request)
+	if err != nil {
+		e := errToAPIError(err)
+		return api.CreateInitiativedefaultJSONResponse{
+			Body:       e,
+			StatusCode: int(e.Code),
+		}, nil
+	}
+	return api.CreateInitiative200JSONResponse(emptySuccess), nil
```
bcspragu, line 22: I don't think you should need any of this `emptysucess` stuff, an empty struct should be sufficient, e.g.

```suggestion
	return api.CreateInitiative200JSONResponse{}, nil
```

cmd/server/pactasrv/initiative.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return api.ListInitiatives200JSONResponse(result), nil
+}
+
+func (s *Server) listInitiatives(ctx context.Context, request api.ListInitiativesRequestObject) ([]api.Initiative, error) {
+	is, err := s.DB.AllInitiatives(s.DB.NoTxn(ctx))
+	if err != nil {
+		return nil, errorInternal(err)
+	}
+	return dereference(mapAll(is, initiativeToOAPI))
```
bcspragu, line 163: Instead of sprinkling `dereference` everywhere, which makes an additional copy of all the elements, why not just make `initiativeToOAPI` return a non-pointer struct?

cmd/server/pactasrv/initiative.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return errorInternal(err)
+	}
+	return nil
+}
+
+// Returns an initiative by ID
+// (GET /initiative/{id})
+func (s *Server) FindInitiativeById(ctx context.Context, request api.FindInitiativeByIdRequestObject) (api.FindInitiativeByIdResponseObject, error) {
+	result, err := s.findInitiativeById(ctx, request)
```
bcspragu, line 121: I understand the exported/unexported structure is to capture all the errors in one place, but I think with the structure I'm proposing, we shouldn't need it. Errors would just be handled explicitly inline

cmd/server/pactasrv/pacta_version.go
```diff
@@ -2,48 +2,164 @@ package pactasrv
...
+		e := errToAPIError(err)
+		return api.UpdatePactaVersiondefaultJSONResponse{
+			Body:       e,
+			StatusCode: int(e.Code),
+		}, nil
+	}
+	return api.UpdatePactaVersion200JSONResponse(emptySuccess), nil
+}
+
+func (s *Server) updatePactaVersion(ctx context.Context, request api.UpdatePactaVersionRequestObject) error {
```
bcspragu, line 97: As an aside, I'm glad we're finally able to have the PATCH-style update semantics we never quite got right in GraphQL, this code is really easy to read + understand

frontend/components/modal/PermissionDenied.vue
```diff
@@ -0,0 +1,37 @@
...
+    sub-header="You aren't authorized to view this page"
+    @closed="navigateBack"
+  >
+    <p>
+      It is likely that you have gotten here either by entering in an invalid
+      URL, or by clicking on a link or button that is not set up correctly.
+    </p>
+    <p>
+      If you think this is a bug (something that you should have access
+      to), or if you want to let us know how you got here, please file a bug.
```
bcspragu, line 29: Instead of `file a bug`, which our European civil servant end-users might not understand, I'd link to a `contact us` page or email alias that we can fill in later

frontend/pages/admin.vue
```diff
@@ -0,0 +1,11 @@
+<script setup lang="ts">
+definePageMeta({
+  middleware: 'admin'
+})
+</script>
+
+<template>
+  <div>
+    <NuxtPage />
+  </div>
```
bcspragu, line 10: I think you can drop the wrapping `div` here. If not, you should be able to use `template` instead

frontend/pages/admin/pacta-version/[id].vue
```diff
@@ -0,0 +1,110 @@
...
+
+const router = useRouter()
+const { pactaClient } = useAPI()
+const { error: { withLoadingAndErrorHandling, handleOAPIError } } = useModal()
+const { fromParams } = useURLParams()
+
+const id = presentOrCheckURL(fromParams('id'))
+
+const prefix = 'admin/pacta-version/[id]'
+const persistedPactaVersion = useState<PactaVersion>('totally different???')
```
Comment (bcspragu) [inline frontend/pages/admin/pacta-version/[id].vue:12]: Probably not the state name we want

frontend/pages/admin/pacta-version/[id].vue
```diff
@@ -0,0 +1,110 @@
...
+)
+const saveChanges = () => withLoadingAndErrorHandling(
+  () => pactaClient.updatePactaVersion(id, changes.value)
+    .then(handleOAPIError)
+    .then(() => { persistedPactaVersion.value = pactaVersion.value })
+    .then(() => router.push('/admin/pacta-version')),
+  `${prefix}.saveChanges`
+)
+
+// TODO(#13) Remove this from the on-mounted hook
```
Comment (bcspragu) [inline frontend/pages/admin/pacta-version/[id].vue:47]: Is the idea that SSR requests aren't currently working? If so, you'll want to update either the `useAPI` composable or underlying plugin to forward the `cookie` header, more or less like we do for AS: https://github.com/Silicon-Ally/AdventureScientists/blob/2b36eafd856949c1e0eaec48c5c07a4004aff092/frontend/plugins/graphql.ts#L62-L68

You'd set it using this: https://github.com/RMI-PACTA/app/blob/7e9e07aacec85441638b33249e75e0481444c54d/frontend/openapi/generated/pacta/PACTAClient.ts#L28

frontend/pages/admin/pacta-version/new.vue
```diff
@@ -17,7 +17,7 @@ const pactaVersion = useState<PactaVersion>(`${prefix}.pactaVersion`, () => ({
 
 const discard = () => router.push('/admin/pacta-version')
 const save = () => withLoadingAndErrorHandling(
-  () => pactaClient.createPactaVersion(pactaVersion.value).then(handleOAPIError),
+  () => pactaClient.createPactaVersion(pactaVersion.value).then(handleOAPIError).then(() => router.push('/admin/pacta-version')),
```
bcspragu, line 20: Do you not want to navigate directly to the new version?

openapi/pacta.yaml
```diff
@@ -222,15 +346,7 @@ paths:
...
-                $ref: '#/components/schemas/User'
-
-        '403':
-          description: caller does not have access or user does not exist 
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/Error'
-
+                $ref: '#/components/schemas/EmptySuccess'
```
bcspragu, line 349: Ah, now I understand the `emptySuccess`. Pretty sure you can just leave this blank (e.g. no `content`) for actual empty success.

cmd/server/pactasrv/conv_oapi_to_pacta.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, fmt.Errorf("pactaVersionCreateToPACTA: nil pointer")
+	}
+	return &pacta.PACTAVersion{
+		Name:        p.Name,
+		Digest:      p.Digest,
+		Description: p.Description,
+	}, nil
+}
+
+func ifNil[T any](t *T, or T) T {
```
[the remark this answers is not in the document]
gbdubs: rofl. Done.

cmd/server/pactasrv/conv_oapi_to_pacta.go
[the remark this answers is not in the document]
gbdubs: The reason I was thinking of doing it this way was so that the `conv` mehtods could access the same set of `errorXYZ` methods + error responses. WDYT?

cmd/server/pactasrv/conv_oapi_to_pacta.go
gbdubs: SGTM

cmd/server/pactasrv/initiative.go
gbdubs: GC done.

cmd/server/pactasrv/initiative.go
gbdubs: GQ - this is because In some contexts a nil input to a conv package should yield a nil result, and sometimes it should yield an error. By defaulting to the more lenient version, then making that requirement explicit in a wrapper method, we make the conv package not include bizlogic about which conversions are expected to always be present.

cmd/server/pactasrv/pacta_version.go
gbdubs: Totally agreed. I feel like in this way (and many others) this project is reaping benefits of the paradigms we didn't enjoy previously.

frontend/components/modal/PermissionDenied.vue
gbdubs: SGTM. Done.

frontend/pages/admin.vue
gbdubs: SG

frontend/pages/admin/pacta-version/[id].vue
Comment (gbdubs) [reply frontend/pages/admin/pacta-version/[id].vue:12]: Lol thanks. I had a very very odd bug where these kept going back into sync.

frontend/pages/admin/pacta-version/[id].vue
Comment (gbdubs) [reply frontend/pages/admin/pacta-version/[id].vue:47]: Yep - that's essentially what I assume is happening. I've copied this suggestion into the bug, and I'll be following up shortly.

frontend/pages/admin/pacta-version/new.vue
gbdubs: Nope - the new version is essentially the page we were on (i.e. just contains the editor). This will bring them back to see their new version in context.

cmd/server/pactasrv/error.go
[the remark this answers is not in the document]
gbdubs: OK I think I've come up with an approach you will find acceptable here! WDYT?

PR #17, opened 2023-09-11
Replace large gif with 10x smaller videos

bcspragu:
This PR removes the gif, since even old browsers support MP4 (still 96kb = 1/5th the size), and it's easier having it all be video instead of half video half `img`

Also, this PR starts using [git lfs](https://git-lfs.com/), which we should be using for large/binary files


gbdubs (approved):
Ayyy thank you!

PR #19, opened 2023-09-11
Implements `findUserByMe` endpoint + trimmings

gbdubs:


cmd/server/pactasrv/conv_pacta_to_oapi.go
```diff
@@ -26,17 +26,32 @@ func initiativeToOAPI(i *pacta.Initiative) (*api.Initiative, error) {
 	}, nil
 }
 
+func userToOAPI(user *pacta.User) (*api.User, error) {
+	if user == nil {
+		return nil, errorInternal(fmt.Errorf("userToOAPI: can't convert nil pointer"))
```
bcspragu, line 31: I hear your point about keeping all the `conv` stuff in the `pactasrv` package so you can access these, but my counterpoint would be that determining the 'type' of an error truly is a concern of the server, and this should just return a `fmt.Errorf(...)` or some `conv`-specific error that the `srv` layer then converts into an `errorWhatever`

Functionally, it probably doesn't matter, just cleaner abstractions and easier testing

bcspragu, line 31: I remember you had a reason for using `*api.Whatever` and `dereference` instead of just using structs directly, but it seems like a lot of contorting to be using pointers, but having to check for errors/disallow them, only to convert them back into not-pointers later

cmd/server/pactasrv/user.go
```diff
@@ -50,3 +53,45 @@ func (s *Server) deleteUser(ctx context.Context, request api.DeleteUserRequestOb
...
+	}
+	// TODO(#18) Handle Multiple Emails in the Token Claims gracefully
+	if len(emails) > 1 {
+		return nil, errorBadRequest("jwt token", "multiple emails in token")
+	}
+	authnID := token.Subject()
+	if authnID == "" {
+		return nil, fmt.Errorf("couldn't find authn id in jwt")
+	}
+	user, err := s.DB.GetOrCreateUserByAuthn(s.DB.NoTxn(ctx), pacta.AuthnMechanism_EmailAndPass, authnID, emails[0])
```
bcspragu, line 92: I would not use a `GetOrCreate` endpoint for something like this. `GetOrCreate`-ing a user should happen on a dedicated `login` or similar endpoint, it should be an error if a user tries to do anything else without an account in the DB.

There's a bunch of reasons to do this, but the main one is to have a single location where we create accounts in the DB.

db/sqldb/user.go
```diff
@@ -52,6 +52,43 @@ func (d *DB) UserByAuthn(tx db.Tx, authnMechanism pacta.AuthnMechanism, authnID
...
+			return nil
+		}
+		if !db.IsNotFound(err) {
+			return fmt.Errorf("looking up user by authn: %w", err)
+		}
+		canonizedEmail, err := pacta.CanonizeEmail(email)
+		if err != nil {
+			return fmt.Errorf("canonizing email: %w", err)
+		}
+		uID, err := d.CreateUser(tx, &pacta.User{
```
bcspragu, line 70: As far as I can tell, this function is only used internally and should probably not be exported.

Similar to above comment, we should really have exactly one public function that can create users - it's a somewhat complicated operation and we don't want multiple places doing it differently.

db/sqldb/user.go
```diff
@@ -52,6 +52,43 @@ func (d *DB) UserByAuthn(tx db.Tx, authnMechanism pacta.AuthnMechanism, authnID
...
+	err := d.RunOrContinueTransaction(tx, func(tx db.Tx) error {
+		u, err := d.userByAuthn(tx, authnMechanism, authnID)
+		if err == nil {
+			user = u
+			return nil
+		}
+		if !db.IsNotFound(err) {
+			return fmt.Errorf("looking up user by authn: %w", err)
+		}
+		canonizedEmail, err := pacta.CanonizeEmail(email)
```
bcspragu, line 66: I'd argue this should happen at the server, not the DB. It's a field, like any other.

deps.bzl
```diff
@@ -543,6 +543,12 @@ def go_dependencies():
         sum = "h1:vV6w1AhK4VMnhBno/TPVCoK9U/LP0PkLCS9tbxHdi/U=",
         version = "v1.1.1",
     )
+    go_repository(
+        name = "com_github_dimuska139_go_email_normalizer",
+        importpath = "github.com/dimuska139/go-email-normalizer",
+        sum = "h1:pJNZnU7uS9MRoYqpoir05B+bCYXrS9sPGE4G1o9EDA8=",
+        version = "v1.2.1",
+    )
```
bcspragu, line 551: I know the open source ecosystem runs on herculean feats of individuals, but I'm still not a fan of random people's not-particularly-widely-used libraries being used in PII sensitive contexts (or really just on the backend in general). If you've audited this and trust it, LGTM, but otherwise, I'd fork it or just make the our own minimal version with the normalizations we care about.

I noted that email normalizing is a fraught endeavor, and I'd argue that farming it out to a random dependency is also a fraught solution.

frontend/components/standard/Nav.vue
```diff
@@ -10,11 +11,11 @@ const menuHidden = useState<boolean>(`${prefix}.menuHidden`, () => false)
...
   return {
     transition: menuHidden.value ? 'max-height .1s ease' : 'max-height .5s ease',
-    overflow: menuHidden.value ? 'hidden' : undefined,
+    overflow: 'hidden',
     'max-height': menuHidden.value ? '0px' : '100vh',
     border: menuHidden.value ? undefined : '2px solid',
-    'margin-top': menuHidden.value ? '0' : '-2px',
-    padding: menuHidden.value ? '0' : '.25rem 0'
+    'margin-top': menuHidden.value ? '0' : '-2px'
+    // padding: menuHidden.value ? '0' : '.25rem 0'
```
bcspragu, line 18: Remove?

frontend/composables/useSession.ts
```diff
@@ -0,0 +1,74 @@
...
+        resolvers.value.push(resolve)
+      })
+    }
+
+    // We're the first to request a user, kick of the request and hop in line at the front of the queue.
+    return new Promise((resolve) => {
+      resolvers.value.push(resolve)
+      void pactaClient.findUserByMe()
+        .then((m) => {
+          console.log(m)
```
bcspragu, line 33: Remove?

frontend/composables/useSession.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (signedIn.value) {
+      await loadCurrentUser()
+    }
+    return {
+      // Will be a Ref with a value of undefined if the user isn't logged in.
+      maybeMe: currentUser,
+      isAdmin,
+      isSuperAdmin
+    }
+  }
```
bcspragu, line 66: As pretty much this whole app should require login, I'd actually just not implement/expose this for now, I've always found this to be kludgy. Do I have an alternative in mind? Definitely not.

openapi/pacta.yaml
```diff
@@ -310,6 +310,31 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/User'
+        '401':
+          description: caller is not logged in - log in to continue 
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EmptySuccess'
```
bcspragu, line 330: `EmptySuccess` Is not what I'd call getting a 401, I'd make a new type (or just leave `content` blank!)

pacta/email.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	normalizer "github.com/dimuska139/go-email-normalizer"
+)
+
+var n = normalizer.NewNormalizer()
+
+func CanonizeEmail(email string) (string, error) {
```
bcspragu, line 14: supernit: `CanonicalizeEmail`?

pacta/email.go
```diff
@@ -0,0 +1,53 @@
...
+	// Performs basic validation, preventing obviously malformed email addresses, but
+	// not capturing the entirety of the RFC 5322 grammar.
+	split := strings.Split(email, "@")
+	if len(split) != 2 {
+		return "", fmt.Errorf("invalid email, wrong number of at-signs: %q", email)
+	}
+	if !isASCII(split[0]) || !utf8.ValidString(split[0]) {
```
bcspragu, line 21: Hold up now: https://en.wikipedia.org/wiki/Email_address#Internationalization

Perfectly reasonable email addresses can have non-ASCII characters in them. Again, thinking of our European users, I'd expect umlauts and other things to appear in addresses, which are valid UTF-8 and valid email addresses by [RFC 6532](https://www.rfc-editor.org/rfc/rfc6532).

If internationalization was done via [punycode](https://en.wikipedia.org/wiki/Punycode), like domain names, then the ASCII check would be fine, but it's just full on Unicode

pacta/email.go
```diff
@@ -0,0 +1,53 @@
...
+	return n.Normalize(email), nil
+}
+
+func isASCII(s string) bool {
+	for _, r := range s {
+		if r > 127 {
+			return false
+		}
+	}
+	fmt.Printf("%s = true\n", s)
```
bcspragu, line 42: Remove, though per the previous comment this whole function should prob go

pacta/email.go
```diff
@@ -0,0 +1,53 @@
...
+	split := strings.Split(email, "@")
+	if len(split) != 2 {
+		return "", fmt.Errorf("invalid email, wrong number of at-signs: %q", email)
+	}
+	if !isASCII(split[0]) || !utf8.ValidString(split[0]) {
+		return "", fmt.Errorf("invalid email, non-ASCII or UTF-8 local part: %q", split[0])
+	}
+	if includesSpace(email) {
+		return "", fmt.Errorf("invalid email, whitespace in email address: %q", split[0])
+	}
```
bcspragu, line 26: I feel like this is one of those all downside, no upside checks and we should just do email verification (which I think we already do with Azure). Since [(quoted) spaces are allowed by the email spec](https://en.wikipedia.org/wiki/Email_address#Local-part), not allowing them is just guaranteeing we make _someone's_ life very annoying for no reason.

scripts/run_db.sh
```diff
@@ -95,6 +95,14 @@ if [ "$SKIP_MIGRATE" = false ]; then
...
-echo "DB is up!"
+echo "
+.__ .  .._.
+[__)|\/| | 
+|  \|  |_|_
+           
+.__ .__    
+|  \[__)   
+|__/[__)   
+"
```
bcspragu, line 106: :grimacing:

bcspragu (approved):

db/sqldb/user.go
gbdubs: SGTM.

deps.bzl
gbdubs: I share your general skepticism, which is why I also audited it. RE: should we create our own fork, I was thinking that Bazel deps provides a sufficient mechanism, since the version + checksum validate that this is the only version that we'll ever use unless we explicitly upgrade it. Is there a way we could more firmly lock in at this version?

frontend/components/standard/Nav.vue
gbdubs: GC

frontend/composables/useSession.ts
gbdubs: GC

frontend/composables/useSession.ts
gbdubs: I think we're going to still need it - examples include public initiatives, and publicly-shared resources.

openapi/pacta.yaml
gbdubs: Fair - previously I couldn't figure out how the return types work with missing `content`! Feel free to change in your upcoming PR to revolutionize errors!

pacta/email.go
gbdubs: TIL! https://wikidiff.com/canonize/canonicalize

Will change.

pacta/email.go
gbdubs: Wait I'm confused - that spec seems to claim that it DOES have to be utf-8 encoded, no?

https://en.wikipedia.org/wiki/International_email#:~:text=The%20set%20of,fully%20internationalized%20email.

pacta/email.go
gbdubs: GC on removing this line ;)

scripts/run_db.sh
gbdubs: Do you have a pointer to the font you use for our other ones? I'd love for them to be consistent.

bcspragu: We link to it! Is called Pagga:

https://github.com/RMI-PACTA/app/blob/2525f7f39e0748935a0f17b1174aa717a469c0da/cmd/server/main.go#L153

deps.bzl
bcspragu: I think code review is the best/easiest tool we have available, the dependency should only change versions if we `go get -u ...` some other dep that requires a newer version, which seems unlikely.

cmd/server/pactasrv/user.go
bcspragu: The other reasons are something like:

1. We shouldn't have exported methods that nobody is using, it increases the package's surface area unnecessarily
2. Given the way our login system works ('get token, send to server'), the only way of account creation that makes sense is `GetOrCreateUser[...]`, we shouldn't make it tempting to create accounts any other way

gbdubs: Wait I might be misunderstanding. When does the token from Azure get sent directly to the server? Aren't we always facilitating that exchange through the Frontend? If so wouldn't just having getOrCreateUser make sense at the FE layer too?

Agreed on the idea of one method at the DB layer. - I'll make `CreateUser` testonly.

db/sqldb/user.go
gbdubs: agreed. fixed.

deps.bzl
gbdubs: SGTM.

openapi/pacta.yaml
gbdubs: I'll leave this to you :)

pacta/email.go
gbdubs: Done.

pacta/email.go
gbdubs: VERY GOOD CATCH. Thank you!

pacta/email.go
gbdubs: SG

scripts/run_db.sh
gbdubs: <3

cmd/server/pactasrv/user.go
bcspragu: Copying image from Slack:
image

bcspragu: The token that would go to a `PACTA` `/login` endpoint would be the RMI-specific token from step 4, and that would get or create the user

PR #20, opened 2023-09-11
Standardizes some Form Inputs + TitleBar Back Button

gbdubs:
This PR attempts to show the general pattern that I'm hoping to use for form validation + error handling in that domain. I'd like for that to be consistent across forms, so if you have ideas on that structure, please let me know.

I also implemented a title-bar-back-button because we're becoming more complex, and it seemed worthwhile.


frontend/components/TitleBar.vue
```diff
@@ -1,19 +1,41 @@
...
-    {{ title }}
-  </h1>
+  <div class="relative flex flex-column gap-2 align-items-start">
+    <PVButton
+      icon="pi pi-angle-left"
+      class="p-button-secondary p-button-text md:hidden p-0 text-600 pr-2"
+      style="margin-left: -.25rem"
+      label="Back"
+      @click="back"
+    />
```
bcspragu, line 26: I can see the value in a 'breadcrumb'-style system that has a list of your current 'depth' into some entity, but less so for just replicating the browser's own "go back" functionality. I'd remove this unless there's some core value prop you have in mind

frontend/components/TitleBar.vue
```diff
@@ -1,19 +1,41 @@
...
+    <PVButton
```
bcspragu, line 27: Wait why are there two buttons?

image

frontend/pages/admin/pacta-version/[id].vue
```diff
@@ -37,6 +39,14 @@ const changes = computed<PactaVersionChanges>(() => {
...
 })
 const hasChanges = computed<boolean>(() => Object.keys(changes.value).length > 0)
+const incompleteFields = computed<string[]>(() => editor.value?.incompleteFields ?? [])
+const isIncomplete = computed(() => incompleteFields.value.length > 0)
+const saveTooltip = computed<string | undefined>(() => {
+  if (!hasChanges.value) { return 'All changes saved' }
+  if (isIncomplete.value) { return `Cannot save with incomplete fields: ${incompleteFields.value.join(', ')}` }
+  return undefined
+})
+const saveDisabled = computed<boolean>(() => saveTooltip.value !== undefined)
```
Comment (bcspragu) [inline frontend/pages/admin/pacta-version/[id].vue:49]: If the idea is that we copy and paste this to each page that has an editor on it, I'm not a fan, as this is a lot of boilerplate that we're going to want to update. I think a system roughly like the following should work and is much clearer + cleaner:

On the page:
```vue
<script setup lang="ts">
// ...
const { newForm } = useForm() // A composable that doesn't currently exist.
const form = newForm('maybeSomeStateLikeKeyThing')
// ...
</script>

<template>
<!-- ... -->

<PactaversionEditor
  :form="form"
  v-model=pactaVersion="pactaVersion"
/>

<!-- ... -->
<div v-tooltip.bottom="form.saveTooltip">
<PVButton
  :disabled="form.saveDisabled" // or name it whatever
  ...
/>
</div>
</template>
```

In the editor component:

```vue
<script setup lang="ts">
const props = defineProps<{
  pactaVersion: PactaVersion
  form: TheFormType
}>()

// Can also specify .length > 0 here, or just let truthy/falsy be decider of validity
props.form.register('Version Name', () => model.value.name)
props.form.register('Docker Image Digest', () => model.value.digest)
</script>
```

Does that make sense? Benefits I see:

1. Most boilerplate now lives in one place in a composable. Makes updating, tweaking, adding functionality easier
2. No need to expose functions on child component, which always felt like a last resort to me.

The only downside I can think of is that the devil is in the details implementation-wise, but I still think it should be pretty straightforward.

bcspragu (commented):
Let's chat about the form strategy when you're back, I think we can come up with something cleaner.

gbdubs:
I folded this into #38.

Issue #10, opened 2023-09-10
Make all error results return non-200 statuses


gbdubs:
Context: https://github.com/RMI-PACTA/app/pull/7/files#r1319227025

PR #21, opened 2023-09-13
Add support for SSR, revamp error handling

### LINKED ISSUE #13: Allow SSR Requests to Fetch Data
We currently get 401's when SSR'ing pages that are requesting data. As a temporary workaround (mostly to avoid complicating an existing PR) I'm using `onMounted` hooks instead.

bcspragu:
This PR gets the SSR ball rolling, such that API requests can now be served via SSR. This is basically just sprinkling `useAsyncData` around and forwarding client cookies during SSR in `useAPI`.

This PR also has a kinda related refactoring of error handling stuff.

## Error Handling Refactor

Now, all error management is done by just `throw`-ing things.

* API returns a non-2xx code? **Throw it** (Axios/the codegen does this for us)
* API response contains an error? **Throw it** (currently in `handleOAPIError`, will go away with #10)
* Error from your `useAsyncData`? **Throw it**
* Data you require is unexpectedly null? **Throw it**

We use [`createError`](https://nuxt.com/docs/getting-started/error-handling#createerror) for everything now, as it gives us a standardized interface to put metadata in, which we can then use to show more useful error messages.

As part of the refactoring, I replaced `NuxtErrorBoundary` with Vue's `onErrorCaptured`, because the former was causing me some problems ([maybe the same problems as others](https://github.com/nuxt/nuxt/issues/15781)). I also moved things from `layouts/default.vue` into `app.vue`, because I feel like error handling is something we'll want everywhere. We can migrate bits back to `layouts/default.vue` as desired.

There's also a new `error.vue` which just wraps our error modal for those times where Nuxt throws a full page error (e.g. a failure during SSR).

Fixes #13

### DIFF
--- a/frontend/components/modal/Error.vue
+++ b/frontend/components/modal/Error.vue
@@ -1,32 +1,27 @@
 <script setup lang="ts">
-import { watch } from 'vue'
-
 interface Props {
-  isFullPage?: boolean
+  routeBackOnClose?: boolean
 }
 
 const props = withDefaults(defineProps<Props>(), {
-  isFullPage: false
+  routeBackOnClose: false
 })
 
 const { error: { errorModalVisible, error: modalError } } = useModal()
 const error = useError()
 const router = useRouter()
 
-watch(errorModalVisible, async (newV, oldV) => {
-  if (!props.isFullPage) {
+const maybeGoBack = async () => {
+  if (!props.routeBackOnClose) {
     return
   }
-  // We only care about the case where the modal was just closed (i.e. has gone from visible -> not visible).
-  if (newV || !oldV) {
-    return
-  }
+
   if (window.history.length > 1) {
     await clearError().then(router.back)
   } else {
     await clearError({ redirect: '/' })
   }
-})
+}
 
 const fullError = computed(() => {
   const err = error.value ?? modalError.value
@@ -49,6 +44,7 @@ const fullError = computed(() => {
     v-model:visible="errorModalVisible"
     header="An error ocurred"
     sub-header="Sorry about that, our team take bug reports seriously, and will try to make it right!"
+    @closed="maybeGoBack"
   >
     <StandardDebug
       label="Error Details"
--- a/frontend/components/modal/Error.vue
+++ b/frontend/components/modal/Error.vue
@@ -1,14 +1,46 @@
 <script setup lang="ts">
-const { error: { errorModalVisible, error } } = useModal()
+import { watch } from 'vue'
+
+interface Props {
+  isFullPage?: boolean
+}
+
+const props = withDefaults(defineProps<Props>(), {
+  isFullPage: false
+})
+
+const { error: { errorModalVisible, error: modalError } } = useModal()
+const error = useError()
+const router = useRouter()
+
+watch(errorModalVisible, async (newV, oldV) => {
+  if (!props.isFullPage) {
+    return
+  }
+  // We only care about the case where the modal was just closed (i.e. has gone from visible -> not visible).
+  if (newV || !oldV) {
+    return
+  }
+  if (window.history.length > 1) {
+    await clearError().then(router.back)
+  } else {
+    await clearError({ redirect: '/' })
+  }
+})
 
 const fullError = computed(() => {
-  return error.value
-    ? {
-        name: error.value.name,
-        message: error.value.message,
-        stack: error.value.stack?.split('\n')
-      }
-    : ''
+  const err = error.value ?? modalError.value
+  if (err instanceof Error) {
+    return {
+      name: err.name ?? '',
+      message: err.message,
+      stack: err.stack?.split('\n')
+    }
+  } else if (err) {
+    return err
+  } else {
+    return ''
+  }
 })
 </script>
 
@@ -19,7 +51,7 @@ const fullError = computed(() => {
     sub-header="Sorry about that, our team take bug reports seriously, and will try to make it right!"
   >
     <StandardDebug
-      label="Error Trace"
+      label="Error Details"
       :value="fullError"
       always
     />
--- a/frontend/components/standard/Debug.vue
+++ b/frontend/components/standard/Debug.vue
@@ -10,25 +10,22 @@ const props = withDefaults(defineProps<Props>(), { always: false, label: 'Techni
 </script>
 
 <template>
-  <!-- TODO(#9) Remove this ClientOnly once the ULS is fixed. -->
-  <ClientOnly>
-    <PVAccordion
-      v-if="showStandardDebug || props.always"
-      class="standard-debug"
+  <PVAccordion
+    v-if="showStandardDebug || props.always"
+    class="standard-debug"
+  >
+    <PVAccordionTab
+      :header="props.label || 'Debug'"
+      content-class="surface-100"
+      header-class="surface-800"
     >
-      <PVAccordionTab
-        :header="props.label || 'Debug'"
-        content-class="surface-100"
-        header-class="surface-800"
+      <div
+        class="code surface-50"
       >
-        <div
-          class="code surface-50"
-        >
-          {{ JSON.stringify(props.value, null, 2) }}
-        </div>
-      </PVAccordionTab>
-    </PVAccordion>
-  </ClientOnly>
+        {{ JSON.stringify(props.value, null, 2) }}
+      </div>
+    </PVAccordionTab>
+  </PVAccordion>
 </template>
 
 <style lang="scss">
--- a/frontend/composables/newModal.ts
+++ b/frontend/composables/newModal.ts
@@ -35,7 +35,7 @@ export const useModal = () => {
     return () => loadingSet.value.delete(loadKey)
   }
   const clearLoading = () => { loadingSet.value.clear() }
-  const withLoadingAndErrorHandling = async <T> (fn: () => Promise<T>, opKey: string): (Promise<T>) => {
+  const withLoading = async <T> (fn: () => Promise<T>, opKey: string): (Promise<T>) => {
     startLoading(opKey)
     const p = fn()
     void p.finally(stopLoading(opKey))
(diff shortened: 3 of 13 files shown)


frontend/components/modal/Error.vue
```diff
@@ -1,14 +1,46 @@
...
+
+const { error: { errorModalVisible, error: modalError } } = useModal()
+const error = useError()
+const router = useRouter()
+
+watch(errorModalVisible, async (newV, oldV) => {
+  if (!props.isFullPage) {
+    return
+  }
+  // We only care about the case where the modal was just closed (i.e. has gone from visible -> not visible).
```
gbdubs, line 20: I try to avoid watch like the plague - since this is event driven I think you can capture this same behavior using the [@closed event](https://github.com/RMI-PACTA/app/blob/2525f7f39e0748935a0f17b1174aa717a469c0da/frontend/components/standard/Modal.vue#L25)

frontend/components/modal/Error.vue
```diff
@@ -1,14 +1,46 @@
 <script setup lang="ts">
-const { error: { errorModalVisible, error } } = useModal()
+import { watch } from 'vue'
+
+interface Props {
+  isFullPage?: boolean
+}
+
+const props = withDefaults(defineProps<Props>(), {
+  isFullPage: false
```
gbdubs, line 9: IIUC, isFullPage is really just "is displaying over the error.vue page" right? If so maybe name it that?

frontend/error.vue
```diff
@@ -0,0 +1,13 @@
...
+// const props = defineProps({
+//   error: Object
+// })
+
+const { error: { errorModalVisible } } = useModal()
+errorModalVisible.value = true
+</script>
+
+<template>
+  <ModalError :is-full-page="true" />
```
gbdubs, line 12: IIRC, `error.vue` doesn't use the default layout. I'd consider asking it to using the approach I used earlier where you move layout.vue into `StandardLayout` so that it can be referenced by both this page + layout.vue, with the contents of the page going in as a `<slot/>`. The value add there would be that the RMI header + footer would still appear in the background, and it would be clearer to the user that they're experiencing an error on the page they were just on, rather than visiting a new site. The downside is that if we cause an error in the NavBar or Footer, we can't recover from it. That _probably_ doesn't matter though because our error-recovery strategies all rely on being able to go back to a page that does include the NavBar and footer (i.e. `router.back` or `router.push('/')`, so I don't think that's a strong reason not to do that.

frontend/composables/newModal.ts
```diff
@@ -52,7 +38,7 @@ export const useModal = () => {
   const withLoadingAndErrorHandling = async <T> (fn: () => Promise<T>, opKey: string): (Promise<T>) => {
```
gbdubs, line 38: This is just `withLoading` now, right? If so let's name it that?

frontend/pages/admin/pacta-version/[id].vue
```diff
@@ -8,9 +8,27 @@ const { fromParams } = useURLParams()
 
 const id = presentOrCheckURL(fromParams('id'))
 
-const prefix = 'admin/pacta-version/[id]'
-const persistedPactaVersion = useState<PactaVersion>(`${prefix}.persistedPactaVersion`)
+const prefix = `admin/pacta-version/${id}`
 const pactaVersion = useState<PactaVersion>(`${prefix}.pactaVersion`)
+const { data: persistedPactaVersion, error, refresh } = await useAsyncData(`${prefix}.getPactaVersion`, () => {
```
Comment (gbdubs) [inline frontend/pages/admin/pacta-version/[id].vue:13]: I like this (generally) but notice that almost all of this is standard/boilerplate.

proposal: can we come up with a wrapper for useAsyncData that does a few things:

1 - reuses the key between the loading OpKey (which we can do now that it isn't used for error reporting) and the useAsyncData key
2 - auto-wraps stuff in `withLoading` so that we don't need to have that ~anywhere
3 - adds the `handleOAPIError` if the return type of the promise can include the `api.Error` type
4 - automatically throws `createError` if an error is returned.
5 - ONLY hands back { data, refresh }, since error handling will all be handled standard.

If we have use cases where we need more customization folks can still choose to use `useAsyncData` in a way that is custom, but the utility described would cover 90% of our use cases with way less boilerplate.

frontend/pages/admin/pacta-version/[id].vue
```diff
@@ -8,9 +8,27 @@ const { fromParams } = useURLParams()
...
+  return withLoadingAndErrorHandling(() => {
+    return pactaClient.findPactaVersionById(id)
+      .then(handleOAPIError)
+  }, `${prefix}.getPactaVersion`)
+})
+if (error.value) {
+  throw createError(error.value)
+}
+if (!persistedPactaVersion.value) {
```
Comment (gbdubs) [inline frontend/pages/admin/pacta-version/[id].vue:22]: We can make this a `present` check, maybe `presentOrCheckURL`? 

Same below on line 28

frontend/pages/admin/pacta-version/[id].vue
```diff
@@ -85,6 +90,7 @@ onMounted(async () => {
         label="Discard Changes"
         icon="pi pi-arrow-left"
         class="p-button-secondary p-button-outlined"
+        to="/admin/pacta-version"
```
Comment (gbdubs) [inline frontend/pages/admin/pacta-version/[id].vue:93]: Thank you!

gbdubs (approved):
LGTM - thanks for doing this!!

frontend/error.vue
bcspragu: My idea was to keep this page as simple as possible for the reasons you mentioned: we don't know what caused the error or where it happened, so the fewer moving parts, the better.

Oddly enough though, halfway through my changes (I don't know what particularly), SSR errors went from having no layout (i.e. just `error.vue`), to having the full layout. I don't know how or why, but I'm going to leave this as is for now, as the behavior matches what you describe here.

frontend/pages/admin/pacta-version/[id].vue
Comment (bcspragu) [reply frontend/pages/admin/pacta-version/[id].vue:13]: Good call, filed #22 to track. Oddly enough, the hardest part about this is the typings. `useAsyncData` has [hideously complicated typings](https://nuxt.com/docs/api/composables/use-async-data#type), and while I can just `Parameters` and `ReturnType` my way to matching them, the whole thing falls apart when you try to actually invoke `useAsyncData` and do stuff with the results.

It'll require some finessing to get it working.

frontend/pages/admin/pacta-version/[id].vue
Comment (bcspragu) [reply frontend/pages/admin/pacta-version/[id].vue:22]: GC, done. I do think we lose relevant context doing so, so I'm going to add an optional third parameter that specifies context.

Update: I went a bit further and merged `present` and the remediation error stuff with Nuxt's `createError`, take another look and let me know what you think. I think standardizing on the Nuxt stuff will make central error reporting (and modal customizing) much easier, and we get to keep strong typing on the error metadata here too

frontend/error.vue
gbdubs: huh - a surprise but a welcome one. Thanks!

frontend/layouts/default.vue
```diff
@@ -1,20 +1,6 @@
...
-const { loading: { onMountedWithLoading, loadingSet }, anyBlockingModalOpen, error: { setError } } = useModal()
-
-const handleError = (event: Event & { reason: Error }) => {
-  event.preventDefault()
-  const { reason } = event
-  setError('fallback')(reason)
-  loadingSet.value.clear()
-}
-
-onMountedWithLoading(() => { /* nothing to do */ }, 'defaultLayout.onMountedWithLoading')
```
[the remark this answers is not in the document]
gbdubs: TYVM

frontend/pages/admin/pacta-version/[id].vue
Comment (gbdubs) [reply frontend/pages/admin/pacta-version/[id].vue:13]: SGTM. I can take a crack at it.

frontend/components/standard/Debug.vue
```diff
@@ -10,25 +10,22 @@ const props = withDefaults(defineProps<Props>(), { always: false, label: 'Techni
 </script>
 
 <template>
-  <!-- TODO(#9) Remove this ClientOnly once the ULS is fixed. -->
```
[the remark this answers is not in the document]
gbdubs: GQ - you probably don't have the value set in local storage, so no hydration mismatch is going to occur. Toggle the little lightning bolt (bottom right) in local mode and try reloading a page that uses `StandardDebug` without `always`

frontend/globalimports/present.ts
```diff
@@ -1,15 +1,23 @@
...
   }
   return t
 }
 
-export const presentOrSuggestReload = <T>(t: T | undefined | null): T => present(t, Remediation.Reload)
-export const presentOrFileBug = <T>(t: T | undefined | null): T => present(t, Remediation.FileBug)
-export const presentOrCheckURL = <T>(t: T | undefined | null): T => present(t, Remediation.CheckUrl)
+export const presentOrSuggestReload = <T>(t: T | undefined | null, cause?: string): T => present(t, Remediation.Reload, cause)
+export const presentOrFileBug = <T>(t: T | undefined | null, cause?: string): T => present(t, Remediation.FileBug, cause)
+export const presentOrCheckURL = <T>(t: T | undefined | null, cause?: string): T => present(t, Remediation.CheckUrl, cause)
```
gbdubs, line 23: looks reasonable to me!

frontend/lib/error/index.ts
```diff
@@ -1,12 +1,32 @@
+import { type NuxtError } from 'nuxt/app'
+
 export enum Remediation {
   None = 'none',
   Reload = 'reload',
   FileBug = 'file-bug',
   CheckUrl = 'check-url',
```
gbdubs, line 7: maybe let's make these more human readable now that we're showing them to humans?

bcspragu: We still aren't showing these to humans, we'll just use them to select the actual thing we display in the error modal

PR #23, opened 2023-09-14
Implement comprehensive error management

### LINKED ISSUE #10: Make all error results return non-200 statuses

bcspragu:
This PR overhauls how we handle errors in our OpenAPI server to be more manageable, and more familiar to [other systems we've used elsewhere](https://github.com/Silicon-Ally/gqlerr).

The basic gist is to only define errors in the OpenAPI spec that are "interesting" or specific to a given endpoint. Things that can occur anywhere (not found, access denied, rate limiting, etc) are defined in the API, but not attached to every single endpoint. These generic errors are thrown as `oapierr.{BadRequest,Internal,NotFound,etc}` errors, with the usual logging metadata attached.

"Specific" errors should be defined on an endpoint, along with the relevant response body, and specified in the response object param (i.e. not an error).

Most of the work here is done in `oapierr`, which more or less mirrors `gqlerr`, minus the GQL-specific stuff naturally. We link into our specific generated code and specific API spec by a caller-specified `oapierr.ResponseConverter` that turns an `oapierr.Error` into an arbitrary JSON-formatted response body. This flexibility means we can turn different errors into different response bodies, using the structs generated directly from our spec.

Now that all errors return non-200 codes, the generated PACTA client will throw an `ApiError` for all failures, see `request.ts` and `catchErrorCodes` in the generated frontend code. `handleOAPIError` is therefore no longer needed and has been removed.

This PR also moves the `conv` things into their own package, which seemed sensible as part of the refactor

Also, the email tests were failing, so I fixed them and then fiddled with them a bit.

Fixes #10


cmd/server/main.go
```diff
@@ -122,8 +123,15 @@ func run(args []string) error {
...
 
 	pactaStrictHandler := oapipacta.NewStrictHandlerWithOptions(srv, nil /* middleware */, oapipacta.StrictHTTPServerOptions{
-		RequestErrorHandlerFunc:  requestErrorHandlerFuncForService(logger, "pacta"),
-		ResponseErrorHandlerFunc: responseErrorHandlerFuncForService(logger, "pacta"),
+		RequestErrorHandlerFunc: requestErrorHandlerFuncForService(logger, "pacta"),
+		ResponseErrorHandlerFunc: oapierr.ErrorHandlerFunc(logger, func(err *oapierr.Error) *oapipacta.Error {
+			// We don't care if it's the default message or not.
+			cm, _ := err.ClientMessage()
+			return &oapipacta.Error{
+				ErrorId: string(err.ErrorID()),
```
gbdubs, line 131: Should we set an error code here?

frontend/openapi/generated/pacta/services/DefaultService.ts
```diff
@@ -71,13 +63,12 @@ export class DefaultService {
...
      * @param id ID of pacta version to delete
-     * @returns EmptySuccess pacta version deleted successfully
-     * @returns Error unexpected error
+     * @returns any pacta version deleted successfully
      * @throws ApiError
      */
     public deletePactaVersion(
         id: string,
-    ): CancelablePromise<EmptySuccess | Error> {
+    ): CancelablePromise<any> {
```
gbdubs, line 71: I get that this is the outcome of us not having an explicit return type for empties, I don't love this though. I'd far prefer we even return `cancelablePromise<void>`. Is that possible? I just really don't want `any`s floating around our codebase.

oapierr/oapierr.go
```diff
@@ -0,0 +1,273 @@
...
+				"received error that was not of type *oapierr.Error",
+				zap.String("type", fmt.Sprintf("%T", err)),
+				zap.Error(err),
+			)
+			e = Internal(err.Error(), zap.Error(err))
+		}
+
+		response := fn(e)
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(e.statusCode)
```
gbdubs, line 243: Ah - discard question above.

openapi/pacta.yaml
```diff
@@ -406,18 +275,27 @@ paths:
...
               
 components:
+  responses:
+    Unauthorized:
+      description: the user is not authorized to access this resource - if logged out, try logging in 
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/Error'
+    Forbidden:
```
gbdubs, line 287: Bad Request, Not Found, Conflict, TooManyRequests?

openapi/pacta.yaml
```diff
@@ -656,19 +534,18 @@ components:
...
         message:
           type: string
-          description: Error message
+          description: Human readable error message (in English)
+        error_id:
+          type: string
+          description: |
+            An enum-like type indicating a more specific type of error.
+
+            TODO: Add an example
```
gbdubs, line 551: If you're revisiting/visiting this question for an example, I think the error for authn for 2+ emails on the token is an excellent candidate.

pacta/email_is_domain.go
```diff
@@ -0,0 +1,63 @@
+package pacta
+
+// isDomainName checks if a string is a presentation-format domain name. This
+// code comes from an unexported function in Go's 'net' package, see
+// https://github.com/golang/go/blob/09300d89e94da7b1f32742e80768acc711b5c590/src/net/dnsclient.go#L75
+func isDomainName(s string) bool {
```
gbdubs, line 6: This feels really hard to maintain. Did you write this, or is it sourced elsewhere?

I really don't love having this in our codebase.

gbdubs (commented):
Overall Looks Good To Me!

cmd/server/main.go
gbdubs: NVM, seen below

PR #24, opened 2023-09-18
Use HTTP 204 No Content for endpoints without response bodies

bcspragu:
This is more semantically accurate and generates the expected `void` response in the frontend client code, see https://github.com/ferdikoomen/openapi-typescript-codegen/issues/607


gbdubs (approved):
nice!

PR #32, opened 2023-10-04
Add basic async task support to PACTA

bcspragu:
Add a new `runner` binary, which will be invoked either locally (via Docker) or remotely (via `aztasks`/Azure Container Apps Jobs). Currently, none of the PACTA-specific handling is here, like passing portfolio information, loading blobs, etc, it's just the basic infra.

Signed-off-by: Brandon Sprague <brandon@sprague.mx>


gbdubs (approved):
I don't see any issues with this and generally understand the approach, but it's a bit too abstract for me to have too many comments right now. I'll do a more thorough review when I start to use this infra more.

bcspragu:
> I don't see any issues with this and generally understand the approach, but it's a bit too abstract for me to have too many comments right now. I'll do a more thorough review when I start to use this infra more.

Sounds good, and happy to chat about the details. Currently, the only missing piece is calling it from some server endpoint, everything else should work from there, both locally (using local docker or Azure Container Apps Jobs) and in deployed environments. It's modelled pretty much exactly on how we do OPGEE runs.

There's also the open question of how data gets transmitted to and from the async portfolio runners. Currently, we transmit run-specific information via env variables (which is fine), but have no way of getting info back. OPGEE does this via blobs in the Azure equivalent of S3, but there are plenty of other schemes PACTA could use (direct DB writes, pubsub, calling the API directly, etc)

PR #34, opened 2023-10-04
Create a new `/blank` page which does what it says on the tin...nothing

bcspragu:
See [1] and [2]

[1] https://github.com/Silicon-Ally/rmi-terraform/pull/6
[2] https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md#redirecturi-considerations

Signed-off-by: Brandon Sprague <brandon@sprague.mx>

### DIFF
--- a/frontend/envs/env.dev
+++ b/frontend/envs/env.dev
@@ -5,4 +5,6 @@ MSAL_USER_FLOW_NAME=B2C_1_susi_dev
 MSAL_USER_FLOW_AUTHORITY=https://rmiauthdev.b2clogin.com/rmiauthdev.onmicrosoft.com/B2C_1_susi_dev
 MSAL_AUTHORITY_DOMAIN=rmiauthdev.b2clogin.com
 MSAL_CLIENT_ID=218f47ee-11b1-459a-b914-b7fa6f107e7b
-MSAL_REDIRECT_URI=https://pacta.dev.rmi.siliconally.dev
+# See https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md#redirecturi-considerations
+MSAL_REDIRECT_URI=https://pacta.dev.rmi.siliconally.dev/blank
+MSAL_LOGOUT_URI=https://pacta.dev.rmi.siliconally.dev/
--- a/frontend/pages/blank.vue
+++ b/frontend/pages/blank.vue
@@ -1,0 +1,11 @@
+<script setup lang="ts">
+// This page is used as a redirect URI for our popup login flow, see
+// https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md#redirecturi-considerations
+
+definePageMeta({
+  layout: 'empty'
+})
+</script>
+
+<template>
+</template>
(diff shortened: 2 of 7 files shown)


frontend/envs/env.dev
```diff
@@ -5,4 +5,6 @@ MSAL_USER_FLOW_NAME=B2C_1_susi_dev
 MSAL_USER_FLOW_AUTHORITY=https://rmiauthdev.b2clogin.com/rmiauthdev.onmicrosoft.com/B2C_1_susi_dev
 MSAL_AUTHORITY_DOMAIN=rmiauthdev.b2clogin.com
 MSAL_CLIENT_ID=218f47ee-11b1-459a-b914-b7fa6f107e7b
-MSAL_REDIRECT_URI=https://pacta.dev.rmi.siliconally.dev
+# See https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md#redirecturi-considerations
+MSAL_REDIRECT_URI=https://pacta.dev.rmi.siliconally.dev/blank
```
gbdubs, line 9: meganit - can we name this something clearer? `blank` has no context, and is more likely to be thought of as `new tab` etc. Very odd ms requires this - let's make the URL clear the purpose, since nothing else on the page will lol. `azure-ad-redirect-destination` ex

frontend/pages/blank.vue
```diff
@@ -0,0 +1,11 @@
+<script setup lang="ts">
+// This page is used as a redirect URI for our popup login flow, see
+// https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md#redirecturi-considerations
+
+definePageMeta({
+  layout: 'empty'
```
gbdubs, line 6: another (maybe simpler) approach? Through an index.html in some folder in the public directory. That way nuxt could never mess this up?

gbdubs (approved):

frontend/envs/env.dev
bcspragu: This is a very reasonable and sensible comment, but I'm going to stick with `/blank` for a few reasons:

1. Pure laziness - I already submitted the TF PR and don't want to deploy both frontends + TF again
2. `/blank` is their recommendation, and I've seen similar things elsewhere. Not sure how strong of a convention it is, but getting a blank page at `/blank` seems reasonable.
3. Only frontend developers associate `new tab` and `blank` (i.e. `target=_blank`)
4. Nobody will ever actually see this, the only time it shows up is for 0.1seconds after signing in, in the Azure popup, just before it closes. The main site window will never show `/blank`

frontend/pages/blank.vue
bcspragu: I did consider it, and that would also be a  totally reasonable approach. I'm going to stick to this for a few reasons:

* The `redirectUri` actually is important in the redirect flow (we currently use the popup flow). We've seen on other projects that there are scenarios (PWAs, some mobile browsers) where popup mode doesn't play nice and you want to fallback to redirect mode, so we're going to want to be able to add little niceties to this page over time, like taking query parameters and doing stuff with them.
* I think there's utility in having the same `<head>` matter that the rest of the site has

I agree there's a minor risk that something like `middleware` causes problems with this page, but since the page is only ever shown for a fraction of a second, I think the risk is minimal.

PR #35, opened 2023-10-11
Overhaul auth management

bcspragu:
Update MSAL integration, mirrors changes we did to OPGEE with some PACTA-specific differences. More specifically:

- **Stay logged in between requests** - Before, if you refreshed the page, you were logged out. Very annoying. This is no longer the case.
- **Cache Azure credentials** - Since the credsrv token + and the Azure ID token expire at the same time, we can use the 'cached' status of the Azure token to decide if we need to refresh our own auth
- **Update MSAL to the latest version**

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,7 +1,8 @@
 {
   "env": {
     "browser": true,
-    "es2021": true
+    "es2021": true,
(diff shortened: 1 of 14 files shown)


gbdubs (approved):
ACTUALLY reviewed :)

PR #39, opened 2023-10-12
Creates Initiative FE

### LINKED ISSUE #22: Make standard composable for data fetching
Copying Grady's [comment from here](https://github.com/RMI-PACTA/app/pull/21/files#r1324605182):

```
I like this (generally) but notice that almost all of this is standard/boilerplate.

proposal: can we come up with a wrapper for useAsyncData that does a few things:

1 - reuses the key between the loading OpKey (which we can do now that it isn't used for error reporting) and the useAsyncData key
2 - auto-wraps stuff in withLoading so that we don't need to have that ~anywhere
3 - adds the handleOAPIError if the return type of the promise can include the api.Error type
4 - automatically throws createError if an error is returned.
5 - ONLY hands back { data, refresh }, since error handling will all be handled standard.

If we have use cases where we need more customization folks can still choose to use useAsyncData in a way that is custom, but the utility described would cover 90% of our use cases with way less boilerplate.
```

gbdubs:
Creates an initiative editor + list view, fixing #37 

- Creates `useSimpleAsyncData` - Fixes #22 
- Adds a Back Button
- Uses purely hierarchical props for forms + sets out a way of doing that consistently.
- Removes Cursive as a fallback, at the request of an unnamed party.
- fixes a typechecking error in `getAccountByHomeId`


frontend/assets/css/theme.css
```diff
@@ -54,19 +54,19 @@ body {
 }
 
 h1 {
-  font-family: Interstate, Roboto, cursive;
+  font-family: Interstate, Roboto, sans-serif;
```
bcspragu, line 57: Music to my eyes, thank you!

frontend/components/ExplicitInputSwitch.vue
```diff
@@ -0,0 +1,24 @@
+<script setup lang="ts">
+interface Props {
+  onLabel: string
+  offLabel: string
+  value: boolean
+}
+const props = defineProps<Props>()
+
+const emit = defineEmits<(e: 'update:value', value: boolean) => void>()
```
bcspragu, line 9: optional super nit: define this in an `Emits` interface for readability, makes it easier to parse at a glance what the contract of a component is when types for both props/emits are at the top.

frontend/components/TitleBar.vue
```diff
@@ -1,19 +1,41 @@
...
 
 <template>
-  <h1 class="inline-block text-primary title-bar">
-    {{ title }}
-  </h1>
+  <div class="relative flex flex-column gap-2 align-items-start">
+    <PVButton
+      icon="pi pi-angle-left"
+      class="p-button-secondary p-button-text md:hidden p-0 text-600 pr-2"
+      style="margin-left: -.25rem"
```
bcspragu, line 23: why not use negative PrimeFlex classes here and below? The classes encourage consistent visual shenanigans. For example, [`-ml-1` exactly matches this style](https://www.primefaces.org/primeflex/margin), AFAICT

frontend/components/form/Field.vue
```diff
@@ -17,13 +17,18 @@ const props = withDefaults(defineProps<Props>(), {
...
   completedLabel: '',
 })
 const slots = useSlots()
 
 const helpTextExists = computed(() => props.helpText !== '' || slots['help-text'] !== undefined)
+const invalidFieldName = computed(() => {
+  if (props.required && !props.completed) { return props.requiredLabel }
+  return undefined
+})
+defineExpose({ invalidFieldName })
```
bcspragu, line 31: I don't actually see this used anywhere, but in either case, here's an example of where I think we can get away without using an `expose`. Instead, I'd run your `invalidFieldName` logic on an event like `blur` and, if invalid, emit an `@invalidField` or whatever event. That way it's a "push" versus a "pull".

frontend/components/initiative/Editor.vue
```diff
@@ -0,0 +1,119 @@
...
+
+const emit = defineEmits<(e: 'update:editorInitiative', ei: EditorInitiative) => void>()
+
+const model = computed({
+  get: () => props.editorInitiative,
+  set: (editorInitiative: EditorInitiative) => { emit('update:editorInitiative', editorInitiative) },
+})
+</script>
+
+<template>
```
bcspragu, line 16: I ran this locally and it looks excellent, great work!

frontend/components/initiative/Editor.vue
```diff
@@ -0,0 +1,119 @@
...
+    >
+      <PVInputText
+        v-model="model.name.currentValue"
+      />
+    </FormField>
+    <FormField
+      label="Initiative ID"
+      help-text="This is the immutable unique identifier for the initiative. It can only contain alphanumeric characters, underscores, and dashes. This value will be shown in URLs, but will typically not be user visible."
+      :required="model.id.isRequired"
+      :completed="isComplete(model.id)"
```
bcspragu, line 32: There's a lot of repetition of these two props on a bunch of `FormField`s, any reason we can't just do something like `:validate="model.id"` and have all the info about "is it required?" and "how do I validate it" baked into the `Editor*`? Then, it's one nice cohesive source of truth and less repetition.

frontend/components/language/Selector.vue
```diff
@@ -0,0 +1,30 @@
+<script setup lang="ts">
+type GenericLang = 'en' | 'de' | 'fr' | 'es'
```
bcspragu, line 2: I'd put this type, and maybe the whole `options` list below somewhere in `@/lib`, I can see this being a thing that gets copied around if not put somewhere meaningful and easily importable.

frontend/components/language/Selector.vue
```diff
@@ -0,0 +1,30 @@
...
+const value = computed<GenericLang>({
+  get: () => props.value,
+  set: (v: GenericLang) => { emits('update:value', v) },
+})
+
+const options: Option[] = [
+  { value: 'en', label: '🇬🇧 English' },
+  { value: 'fr', label: '🇫🇷 Français' },
+  { value: 'de', label: '🇩🇪 Deutsch' },
+  { value: 'es', label: '🇪🇸 Español' },
```
bcspragu, line 19: optional: instead of putting the emoji in the label and trusting the user's browser to emojify it correctly, you could use a dropdown template. In fact, [the PrimeVue example for custom templates uses flags](https://primevue.org/dropdown/#template). They do this crazy thing where they use [a single, loooong image](https://primefaces.org/cdn/primevue/images/flag/flags_responsive.png) with all the flags and then basically index into it with `background-image`.

uhh, or emojis in the label might be fine too :shrug:

frontend/components/pactaversion/Selector.vue
```diff
@@ -0,0 +1,37 @@
...
+const props = defineProps<Props>()
+const emits = defineEmits<(e: 'update:value', value: string | undefined) => void>()
+const value = computed({
+  get: () => props.value,
+  set: (v: string | undefined) => { emits('update:value', v) },
+})
+
+const pactaClient = await usePACTA()
+const prefix = 'components/pactaversion/Selector'
+const { data: pactaVersions, refresh } = await useSimpleAsyncData(
```
bcspragu, line 14: Loading data in components is one of those _kinda_ anti-patterns we've done in the past that I'd like us to move away from. Passing in `pactaVersions` as a prop and loading it in the relevant page means we don't have a dozen components loading semi-overlapping data on page load.

This is especially important since Nuxt/your useSimpleAsyncData doesn't do caching. So, for example, N `PactaversionSelector`s on a page would issue N `pactaClient.listPactaVersions` calls

frontend/composables/useMSAL.ts
```diff
@@ -236,7 +236,7 @@ export const useMSAL = async () => {
     const request: SilentRequest = {
       scopes,
       forceRefresh: false, // Set this to "true" to skip a cached token and go to the server to get a new token
-      account: inst.getAccount({ homeAccountId: account.value.homeAccountId }) ?? undefined,
+      account: inst.getAccountByHomeId(account.value.homeAccountId) ?? undefined,
```
bcspragu, line 239: [`getAccountByHomeId` is deprecated](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/0f5e6c2142889ea9f43d808f270aec9264afd363/lib/msal-browser/src/app/PublicClientApplication.ts#L207), which is why I changed it. If you're seeing otherwise, you might not have run `npm install` after pulling in my most recent changes, which updated the MSAL library.

frontend/composables/useSimpleAsyncData.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const useSimpleAsyncData = async <T>(key: string, fn: () => Promise<T>): Promise<SimpleAsyncDataReturn<T>> => {
+  const { loading: { withLoading } } = useModal()
+  const fnWithLoading = () => withLoading(fn, key)
+  const { data: dataRef, refresh, error } = await useAsyncData(key, fnWithLoading)
+  if (error.value) {
+    throw createError(error.value)
+  }
+  const data = dataRef as unknown as Ref<T>
```
bcspragu, line 13: Lol I remember saying "the trickiest part is getting the typings right", which I see borne out here.

More seriously though, are we confident that we're not losing something in translation here? I can see the "eliding via `unknown`" as causing difficult to debug typing issues.

Part of that is mitigated because you don't implement the full `useAsyncData` interface, which has things like `transform` and `pick` that complicate the types (and indeed, the actual return values) quite a bit.

frontend/lib/editor/common.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface ForUse<T> {
+  isRequired: boolean
+  isCompleted: ComputedRef<boolean>
+  value: WritableComputedRef<T>
+}
+
+export enum Validation {
+  NotEmpty = 'NotEmpty',
+  AlphanumericAndDashesAndUnderscores = 'AlphanumericAndDashesAndUnderscores',
+}
```
bcspragu, line 12: This all seems like a well-structured library, though I worry we're reinventing the wheel along a few different axes here:

- These kind of form validations are built into HTML5, see https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#using_built-in_form_validation
- AFAICT, our requirements here are not unique and we could probably benefit from an off-the-shelf solution here.
	- For example, [PrimeVue has integration with VeeValidate](https://primevue.org/inputtext/#form), a popular Vue validation library (which should work fine with Nuxt)

frontend/lib/editor/common.ts
```diff
@@ -0,0 +1,61 @@
...
+  const result: Partial<R> = {}
+  if (field.originalValue === field.currentValue) {
+    return result
+  }
+  result[field.name] = field.currentValue
+  return result
+}
+
+export const asValue = <R, N extends keyof R>(field: EditorField<R, N>): Pick<R, N> => {
+  return { [field.name]: field.currentValue } as unknown as Pick<R, N>
```
bcspragu, line 33: I spent a few minutes poking around with this. I'm still convinced that it is definitely doable in TypeScript, even if a computer said otherwise. The best I could do is just removing the `unknown` though ([playground link](https://www.typescriptlang.org/play?target=99&jsx=0#code/FAUwHgDg9gTgLgAhAOwK4FsEDUCGAbASwBMc4CpkEBvYBBAOSjgFF0I4BPBAXgQHJGLNpz4AaWggCCeCAAscadCBgEAxpOREAIjgDOskLo1EAqpuW7VsQz37S5CjMrXGd+w8bNELVmIbHAAL7AoJCwiATIcMoAZjiqIAjMRARwsABiBCB4RAA8AEqiDDhKSGDRmroIANYgHFAxCPkAfNQSyCUgAFzFShJ4OABG2T26cCrIAOYSBLr5IACOqAR+RAD8PYNQUHggChIAbvjEpOTIG9jHJGQUCAA+CKjmMZEgRBKwBJOR+Lh4qN0mgBteidAC6ElUqBgfiifwBPXyIPBQRC4Gg8AQVmQYwQenhiV4BSK9DKFSIVVq9UaLQAFC9skQeslUhksjliQxmgBKHoABTU1U59Fa3FaNDofjg0MoVAQQIZOQAdB0lGCeoqiEqoTCUHACQhAniqgLVELClzUUA)), that seemed to work.

```suggestion
  return { [field.name]: field.currentValue } as Pick<R, N>
```

frontend/lib/editor/common.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return []
+}
+
+export const asComputed = <R, N extends keyof R>(field: WritableComputedRef<EditorField<R, N>>): WritableComputedRef<R[N]> => {
+  return computed<R[N]>({
+    get: () => field.value.currentValue,
+    set: (v: R[N]) => { field.value.currentValue = v },
+  })
+}
```
bcspragu, line 48: When would this function be necessary? Doesn't look like it's currently used.

frontend/lib/editor/initiative.ts
```diff
@@ -0,0 +1,174 @@
...
+      currentValue: initiative.pactaVersion,
+    },
+    createdAt: {
+      name: 'createdAt',
+      label: 'Created At',
+      originalValue: initiative.createdAt,
+      currentValue: initiative.createdAt,
+    },
+  }
+}
```
bcspragu, line 98: I like this part, in the sense that it describes each field and required/validation status, but everything else below is a ton of very error-prone boiler plate. Some ideas:

- Take a look at VeeValidate and see if it's worth using that (or at least borrowing some ideas)
- If not, it _should_ be possible to replace `computedX`, `getComputedXChanges`, and `getComputedIncompleteFields` with just one of each, instead of one-per-entity, by iterating over the object properties and doing more or less what you're doing now.

frontend/pages/admin/initiative/[id].vue
```diff
@@ -0,0 +1,78 @@
...
+        />
+      </div>
+    </div>
+    <StandardDebug
+      :value="editorInitiative"
+      label="Editor Initiative"
+    />
+    <StandardDebug
+      :value="changes"
+      label="PV Changes"
```
Comment (bcspragu) [inline frontend/pages/admin/initiative/[id].vue:75]: `InitiativeChanges`?

frontend/pages/admin/initiative/new.vue
```diff
@@ -0,0 +1,74 @@
...
+} = initiativeEditor({
+  id: '',
+  name: '',
+  affiliation: '',
+  publicDescription: '',
+  internalDescription: '',
+  requiresInvitationToJoin: false,
+  isAcceptingNewMembers: false,
+  isAcceptingNewPortfolios: false,
+  language: Initiative.language.EN,
```
bcspragu, line 25: Should there be a TODO here to set this based on the user's chosen language?

bcspragu (approved):
Looks great, solid base to build the rest of the app on. Main concern is that the form stuff seems really overdone, I'd like to better understand how existing solutions handle stuff like this.

frontend/components/TitleBar.vue
gbdubs: TIL!!! Thank you! I had no idea they had negative margins in primeflex! Thanks!

frontend/components/form/Field.vue
gbdubs: Ah yes! It's no longer nescessary - it was the anti-pattern I was trying to design out. I'll leave this open to discuss the other part of the suggesiton.

frontend/components/initiative/Editor.vue
gbdubs: Thanks!

frontend/components/initiative/Editor.vue
gbdubs: Great Idea! Done, after much struggle. This PR is WAY BETTER OFF for this comment.

frontend/components/language/Selector.vue
gbdubs: Good idea, done, in a way that is reusable for the next PR :)

frontend/components/language/Selector.vue
gbdubs: https://answers.microsoft.com/en-us/windows/forum/all/flag-emoji/85b163bc-786a-4918-9042-763ccf4b6c05?page=4#:~:text=LUDICROUS%2DREDICULOUS%2DDISGUSTING%2DPATHETIC

Yep we'll use images. Done.

frontend/composables/useMSAL.ts
gbdubs: Ahhh yes that's it. I was getting TSC failures.

frontend/lib/editor/common.ts
gbdubs: This is a fair criticism. Since this has changed quite a bit, and I think simplified quite a bit, would you take another look before I go too far into the Vee-Validate rabbit hole? 

Context: I've messed around with validation libraries in the past, and my prior experiences with them (most recently in abound) have always been poor - they do that thing we always talk about of offering leaky abstractions that aren't actually solving for the complexity they try to tackle.

frontend/lib/editor/common.ts
gbdubs: Thoroughly improved! PTAL.

frontend/lib/editor/common.ts
gbdubs: Good question - it is used - it's for creates, which take in `R` rather than a `Partial<R>`.

frontend/lib/editor/initiative.ts
gbdubs: Done the second one - PTAL, and lmk if you're comfortable.

frontend/pages/admin/initiative/[id].vue
Comment (gbdubs) [reply frontend/pages/admin/initiative/[id].vue:75]: GC

frontend/pages/admin/initiative/new.vue
gbdubs: GQ - the requirements specified english as the default.

frontend/composables/useSimpleAsyncData.ts
gbdubs: Good call - the unknown isn't nescessary, so I removed it.

frontend/lib/editor/initiative.ts
```diff
@@ -97,78 +80,16 @@ const createEditorInitiative = (initiative: Initiative): EditorInitiative => {
...
-    editorInitiative: ei,
-    setInitiative,
-    incompleteFields,
-    changes,
-    hasChanges,
-    isIncomplete,
-    initiative,
-  }
-}
+export type ExampleEditor = EditorFieldsFor<Example>
```
bcspragu, line 95: What's the purpose of this?

frontend/public/img/flags/es.svg
bcspragu, line 1: nit: This is a beautiful flag, but we're going to render it at most like 22px across, and it's nearly 100KB. With only ~4 languages, this isn't a real issue, but these are totally overkill, esp since they'll get rendered and loaded on every page

bcspragu (approved):
The updates to the editor/validation are great, much more succinct/less boilerplate. The only problem is that now the code is completely incomprehensible to me and will likely be hard to update, but I guess that's the tradeoff for this sort of metaprogramming, so LGTM

frontend/lib/editor/initiative.ts
gbdubs: GQ - it's used as a prop input type to `InitiativeEditor.vue`.

gbdubs: Oh nvm. I see what you mean. Removed. I was playing around with this to understand how ` | undefined` differed from `var?:`

frontend/public/img/flags/es.svg
gbdubs: Great call - replaced with a much smaller flag.

gbdubs: 40kb. Bad, but not terrible

PR #40, opened 2023-10-12
Creates i18n infrastructure

gbdubs:
- Adds `vue-i18n` and `nuxt-i18n`, configured such that:
    - we'll have the locale in the url (`/fr/admin`) etc. This is easy to change if RMI wants to later.
    - we store the user's preference in a cookie, and that cookie works (at least locally) to deliver an SSR'ed i18n page.
    - we guess the user's location based on `accept-lang` and a few other heuristics, but honor a user's choice once made.
    - we offer english pages without prefix (`/admin` and `/en/admin` go to the same place)
    - the language files are lazy-loaded, so only current language gets pulled in by using `VueI18nVitePlugin` - you can see in the network tab when we switch langauges we fetch the `lang/*.json?import` file.
    - configures the fallback to be English, with the expectation that we won't actually write any of the other-language strings until the pre-launch stage of the project.
- Creates a pattern that I hope we use everywhere for i18n.
    - uses `objectNotation: true` to allow `Prefix.Suffix` to be a nested property in JSON.
    - Groups language strings by file `prefix` - seems like an appropriate thing for now.
    - uses `prefix` with `t` in the template to create a helper `tt` to avoid having to write prefixes.
    - Generally uses the full string as the translation key, but uses a placeholder for longer strings.
    - Implements this pattern on the Nav + Footer + a few other places so you can see what it looks like. 
- Re-configures the footer to support a language selector. I decided on the footer over the header because as a I looked around and most functional websites put the selector in the footer, while most informational/one-time websites put it in the header.
- I went with flags in the selector mostly for visibility/legibility. Without flags it's hard to draw the eye to the selector. I don't love it for a bunch of colonial reasons, but I think given how out-of-view it is after the first selection, I think it's probably a reasonable tradeoff.
- Creates a one-time toast to alert folks to the fact that we support multiple languages.


frontend/components/locale/Selector.vue
```diff
@@ -0,0 +1,106 @@
...
+const { languageWasSelectedOrDismissed } = useLocalStorage()
+
+// We go through this rigamarole because the additional information passed into
+// the i18n configuration is present here (namely, flag + name), but it doesn't
+// appear in the typescript configuration.
+interface Option {
+  code: string
+  flag: string
+  name: string
+}
```
bcspragu, line 17: I might be misunderstanding, but I think you can drop the `computed`s and do:

```ts
const i18n = useI18n();
const locales = i18n.locales as Option[];
const locale = i18n.locale as Option;
const { t } = i18n; 

if (!locales.value.includes(locale.value)) {
  throw createErrorWithRemediation('locale not found', Remediation.FileBug)
}

// Can just use `locale` from here.
```

frontend/lang/en.json
```diff
@@ -0,0 +1,47 @@
...
+        "Loading": "Loading..."
+    },
+    "LocaleSelector": {
+        "Availabile In": "This site is available in:",
+        "Bottom Right": "You can change the language at any time in the bottom right corner of the page. ↘️"
+    },
+    "ModalError": {
+        "Heading": "An error occurred",
+        "Subheading": "Sorry about that, our team take bug reports seriously, and will try to make it right!",
+        "Error Details": "Error Details",
```
bcspragu, line 13: super nit: trailing whitespace, maybe tweak our lint to catch this?

frontend/lang/en.json
```diff
@@ -0,0 +1,47 @@
...
+    },
+    "LocaleSelector": {
+        "Availabile In": "This site is available in:",
+        "Bottom Right": "You can change the language at any time in the bottom right corner of the page. ↘️"
+    },
+    "ModalError": {
+        "Heading": "An error occurred",
+        "Subheading": "Sorry about that, our team take bug reports seriously, and will try to make it right!",
+        "Error Details": "Error Details", 
+        "Common Steps": "",
```
bcspragu, line 14: Should this have some copy? Won't it just be blank otherwise? On the other hand, I don't see it used anywhere.

frontend/layouts/default.vue
```diff
@@ -13,7 +15,7 @@ onMountedWithLoading(() => { /* nothing to do */ }, 'defaultLayout.onMountedWith
     >
       <main
         class="px-3 md:px-4 w-full lg:w-10 xl:w-8 mx-auto"
-        style="min-height: calc(100vh - 9.25rem - 4px);"
+        style="min-height: calc(100vh - 10.5rem - 4px);"
```
bcspragu, line 18: Should this calc be pulled out into a variable somewhere? Seems error prone to remember to update it here and `StandardContent`

frontend/lib/error/index.ts
```diff
@@ -19,6 +19,7 @@ interface PACTAError extends NuxtError {
 }
 
 export const createErrorWithRemediation = (err: string | Partial<NuxtError>, r: Remediation): PACTAError => {
+  console.log(err)
```
bcspragu, line 22: remove? This should get logged up the stack if it's thrown correctly

frontend/nuxt.config.ts
```diff
@@ -1,8 +1,11 @@
 // https://nuxt.com/docs/api/configuration/nuxt-config
 export default defineNuxtConfig({
+  modules: [
+    '@nuxtjs/i18n',
+  ],
   build: {
     // https://primevue.org/installation/#nuxtintegration
-    transpile: ['primevue'],
+    transpile: ['primevue', 'vue-i18n'],
```
bcspragu, line 8:
```suggestion
    
    transpile: ['primevue'],
    transpile: [
      'primevue', // https://primevue.org/installation/#nuxtintegration
      'vue-i18n',
	],
```

frontend/nuxt.config.ts
```diff
@@ -41,4 +44,24 @@ export default defineNuxtConfig({
     ],
     dirs: ['globalimports'],
   },
+  i18n: {
+    // TODO(grady) Set a Base URL once we have it for SEO https://i18n.nuxtjs.org/guide/seo#:~:text=You%20must%20also%20set%20the%20baseUrl%20option%20to%20your%20production%20domain%20in%20order%20to%20make%20alternate%20URLs%20fully%2Dqualified%3A
+    // baseUrl: 'https://my-nuxt-app.com'
```
bcspragu, line 49: But we do have a base URL! In dev, it's `https://pacta.dev.rmi.siliconally.dev`. Seems like this should be populated from `envs/` like other env-specific config

frontend/package.json
```diff
@@ -42,6 +43,7 @@
     "primeflex": "^3.3.1",
     "primeicons": "^6.0.1",
     "primevue": "^3.32.0",
-    "uuid": "^9.0.0"
+    "uuid": "^9.0.0",
+    "vue-i18n": "^9.5.0"
```
bcspragu, line 47: Do you need to specify `vue-i18n` here explicitly, or it it sufficient to let `@nuxtjs/i18n` pick the version implicitly? I ask because specifying overly specific dependencies can cause versioning issues, so if this is just a transient for us, might as well drop it.

bcspragu (approved):
Looks great, main general comment is that not all links have been updated [1], so you occassionally just get dumped back to the default language.

Is there a scheme (within vue/nuxt-i18n) we can use that's a little more bulletproof? It's easy to forget `localePath` when adding links/buttons/general nav

[1] `/pages/admin/index.vue` and `/pages/admin/pacta-version/index.vue` were the two I noticed

frontend/components/locale/Selector.vue
gbdubs: Good question - the `locale` from `i18n` is JUST the `code` - it's a string.

frontend/lang/en.json
gbdubs: Spent 1/2 hr trying to get Json linting working, to no avail. I'm going to punt on this for now.

frontend/lang/en.json
gbdubs: Good catch - fixed.

frontend/layouts/default.vue
gbdubs: GC - done.

frontend/lib/error/index.ts
gbdubs: These aren't showing up in logs, and it's proving challenging to debug stack traces without them.

frontend/nuxt.config.ts
gbdubs: GC

frontend/nuxt.config.ts
gbdubs: GC fixed

frontend/package.json
gbdubs: Yep - we directly import from it in some of the code already :)

gbdubs:
> Looks great, main general comment is that not all links have been updated [1], so you occassionally just get dumped back to the default language.
> 
> Is there a scheme (within vue/nuxt-i18n) we can use that's a little more bulletproof? It's easy to forget `localePath` when adding links/buttons/general nav
> 
> [1] `/pages/admin/index.vue` and `/pages/admin/pacta-version/index.vue` were the two I noticed

Great question + suggestion. I ended up creating some middleware that tests for this, and logs in the console when we fail to meet it. We can beef that up if/when we need to. That (in tandem with making the prefix consistent) will make these navigation errors a lot easier to spot and remediate, though I wish we could do something more compiley.

PR #41, opened 2023-10-16
Frontend Build Out

gbdubs:
NOTE: This PR is a poorly structured amalgomation of a bunch of small changes. In the future I'll try to do better to structure changes like these as smaller, more focused PRs - just during this initial build out, so much is changing so quickly, I'm struggling to find the "right" boundaries for increments of work.

Contentious
- **Starts Populating UserID in context**, and in doing so, moves the logic for extracting a PACTA User ID from a Authn User ID and moves it into a server handler in the `main.go` method. I don't love this so if you have better Ideas on how this can be done, please let me know.
- Creates a **Fake Users Modal** (since MSAL doesn't have a direct authn mechanism that I can find, I think this kind of "shared test account" strategy may be the best we can do.
- merges `useSession.signedIn` and `useMSAL.isAuthenticated` to fix a few issues with hydration mismtaches/SSR 500s.

Non-Contentious
- **Creates OpenAPI methods for**
    - User Editing + Retrieval
    - Initiative Invitation Editing + Claiming
    - Initiative Membership Listing
- Creates **rudimentary** frontends for initiatives, users, relationships, etc. The goal of these isn't to be a polished final result, but demonstrate basic functionality. I've added i18n everywhere I expect to be stable/permanent, but have not bothered with it elsewhere.


cmd/server/main.go
```diff
@@ -340,6 +342,61 @@ func rateLimitMiddleware(maxReq int, windowLength time.Duration) func(http.Handl
...
+		authnID := token.Subject()
+		if authnID == "" {
+			return nil, fmt.Errorf("couldn't find authn id in jwt")
+		}
+		user, err := db.GetOrCreateUserByAuthn(db.NoTxn(c), pacta.AuthnMechanism_EmailAndPass, authnID, email, canonical)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get user by authn: %w", err)
+		}
+		return jwtauth.WithUserId(c, string(user.ID)), nil
+	}
```
bcspragu, line 383: Discussed offline:

* Move this into a package/struct, like `session.WithAuthn`
* Use a 'read-only' DB function (the `GetOrCreate` should only be used on login)
* Use `session.{WithUserID,UserIDFromContext}` for setting/retrieving the ID

cmd/server/main.go
```diff
@@ -340,6 +342,61 @@ func rateLimitMiddleware(maxReq int, windowLength time.Duration) func(http.Handl
...
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx, err := fn(r.Context())
```
bcspragu, line 387: nit: I think this will go away with the above comment, but just in case it doesn't: normally, one only returns a new context from functions that are dedicated to contexts, e.g. `WithX` or `context.WithX` functions. So I'd change it to something like:

```go
fn := func(c context.Context) (pacta.UserID, error) { ... }
[...]
return func(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		uID, err := fn(r.Context())
		ctx := session.WithUserID(ctx, uID)
		next.ServeHTTP(w, r.WithContext(ctx))
	}
}
```

cmd/server/pactasrv/initiative_invitation.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	result, err := conv.InitiativeInvitationToOAPI(ii)
+	if err != nil {
+		return nil, err
+	}
+	return api.GetInitiativeInvitation200JSONResponse(*result), nil
+}
+
+// Claims this initiative invitation, if it exists
+// (POST /initiative-invitation/{id})
```
bcspragu, line 65: nit: I think of "claiming an invitation" as a non-CRUD action that should get a custom verb, like `POST /initiative-invitation/{id}:claim`. Benefits of this:

- Unambigious about what it's for, makes reading logs and stuff easier
- If you have another "updating an initiative invitation" endpoint you need to add, you won't have a collision on `POST /initiative-invitation/{id}`

[Some Google Cloud guide that discusses this at length](https://cloud.google.com/apis/design/custom_methods)

cmd/server/pactasrv/initiative_invitation.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+	err = s.DB.Transactional(ctx, func(tx db.Tx) error {
+		ii, err := s.DB.InitiativeInvitation(tx, pacta.InitiativeInvitationID(request.Id))
+		if err != nil {
+			return fmt.Errorf("looking up initiative invite: %w", err)
+		}
+		if !ii.UsedAt.IsZero() || ii.UsedBy != nil {
+			if ii.UsedBy != nil && ii.UsedBy.ID == userID {
```
bcspragu, line 77: My usual grumbling about new paradigms I don't like: the whole "We use objects for everything and only populate the IDs" makes all of this logic one conditional comparision more than it otherwise would need to be. E.g. as compared to

```go
if ii.UsedByID == userID { ... }
```

cmd/server/pactasrv/initiative_invitation.go
```diff
@@ -0,0 +1,118 @@
...
+				// We don't return an error if the same user tries to claim the same invitation twice,
+				// which might happen by accident, but wouldn't impact the state of the initiative memberships.
+				// We may want to log this, though.
+				return nil
+			} else {
+				return fmt.Errorf("initiative is already used: %+v", ii)
```
bcspragu, line 83: We should be using oapierr.BadRequest (or similar) for this. For this case in particular, we likely want a custom error code so we can show a specific error to the end user. Then you'd just need to update the error logic below to:

```go
if errors.Is(err, oapierr.Error{}) { // or something like that, could also make oapierr.Is(err) or similar
	return err
} else if err != nil {
	return oapierr.Internal(...)
}
```

You could even make a lil helper, like `return oapierr.OrInternal(err)`

cmd/server/pactasrv/initiative_invitation.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil
+	})
+	if err != nil {
+		return nil, oapierr.Internal("failed to claim initiative invitation", zap.Error(err))
+	}
+	return api.ClaimInitiativeInvitation204Response{}, nil
+}
+
+// Returns all initiative invitations associated with the initiative
+// (GET /initiative/{id}/invitations)
```
bcspragu, line 106: Why is this `GET /initiative/{id}/invitations` and not `GET /initiative-invitation`? I think the standard RESTful approach would be the latter, [ex](https://cloud.google.com/apis/design/standard_methods)

cmd/server/pactasrv/initiative_user_relationship.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	result, err := dereference(mapAll(iurs, conv.InitiativeUserRelationshipToOAPI))
+	if err != nil {
+		return nil, err
+	}
+	return api.ListInitiativeUserRelationshipsByUser200JSONResponse(result), nil
+}
+
+// Returns all initiative user relationships for the initiative that the user has access to view
+// (GET /initiative/user-relationships/{id})
```
bcspragu, line 30: This URL scheme is not intuitive to me, e.g. the difference between:

- `GET /initiative/{id}/user-relationships`, and
- `GET /initiative/user-relationships/{id}`, and

The benefit of RESTful stuff is that usually it's pretty obvious from the URL what the endpoint does, but I'm not getting that from these. How about something like:

- `GET /initiative/{id}` - Gets the initiative and the relationships, can optionally hide the latter behind a query param like `withRelationships=true`
	- If for some reason we're not frequently loading initiatives + their relationships at the same time, could also do `/initiative/{id}/relationships`, basically like it is now.
- `GET /user/{id}/initiatives` - Gets a list of initiatives the user has access to

cmd/server/pactasrv/initiative_user_relationship.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	result, err := dereference(mapAll(iurs, conv.InitiativeUserRelationshipToOAPI))
+	if err != nil {
+		return nil, err
+	}
+	return api.ListInitiativeUserRelationshipsByInitiative200JSONResponse(result), nil
+}
+
+// Returns the initiative user relationship from this id, if it exists
+// (GET /initiative/{initiativeId}/user-relationship/{userId})
```
bcspragu, line 45: This API surface seems really broad to me, what's the motivation for having three different endpoints that return a user's relationship to an initiative? Unless we expect the list of relationships to be exceptionally large (>1,000), I'd just let the client filter for the stuff they're looking for.

More API surface means more stuff to secure, and more DB functions/tests/code to maintain overall.

cmd/server/pactasrv/pactasrv.go
```diff
@@ -100,3 +103,11 @@ func dereference[T any](ts []*T, e error) ([]T, error) {
 	}
 	return result, nil
 }
+
+func getUserID(ctx context.Context) (pacta.UserID, error) {
+	userID, err := jwtauth.UserIDFromContext(ctx)
```
bcspragu, line 108: CI seems to be passing, but I have no idea how this code works, or the `jwtauth.WithUserId` elsewhere. These functions don't appear to exist in [the `jwtauth` package](https://pkg.go.dev/github.com/go-chi/jwtauth/v5#section-readme)

cmd/server/pactasrv/user.go
```diff
@@ -4,72 +4,83 @@ import (
...
-		return nil, oapierr.Internal("couldn't parse email claim in token")
-	}
-	// TODO(#18) Handle Multiple Emails in the Token Claims gracefully
-	if len(emails) > 1 {
-		return nil, oapierr.BadRequest("multiple emails in token")
-	}
-	email := emails[0]
-	canonical, err := pacta.CanonicalizeEmail(email)
+	// Looking for how users get created or populated into the context?
+	// It's in the HTTP handler for adding a user to the context in main.go
```
bcspragu, line 72: After you move everything into a `session` package, it'll be more obvious and this comment can be removed.

frontend/components/LinkButton.vue
```diff
@@ -83,21 +95,30 @@ const disabled = computed(() => {
...
+    'p-button-loading': !!props.loading,
     'p-button-loading-label-only': props.loading !== undefined && props.icon === undefined && props.label !== undefined,
     'no-underline': true,
     'click-does-nothing': disabled.value,
   }
+  if (isActive.value) {
+    result[props.activeClass] = true
+  } else {
+    result[props.inactiveClass] = true
+  }
```
bcspragu, line 120: I think you can just put these directly into the map above, like:

```ts
return {
  // ... other stuff ...
  `${props.activeClass}`: isActive.value,
  `${props.inactiveClass}`: !isActive.value,
}
```

frontend/components/initiative/Toolbar.vue
```diff
@@ -0,0 +1,76 @@
...
+
+interface Props {
+  initiativeId: string
+  initiativeUserRelationships: InitiativeUserRelationship[]
+}
+const props = defineProps<Props>()
+
+const isManager = computed(() => {
+  const mm = maybeMe.value
+  return (!!mm && props.initiativeUserRelationships.some(r => r.manager && r.userId === mm.id)) || true
```
bcspragu, line 20: Lol I'm guessing the `|| true` was for testing? Remove pls

frontend/components/initiative/Toolbar.vue
```diff
@@ -0,0 +1,76 @@
...
+})
+const isMember = computed(() => {
+  const mm = maybeMe.value
+  return !!mm && props.initiativeUserRelationships.some(r => r.member && r.userId === mm.id)
+})
+const canEdit = computed<boolean>(() => isManager.value || isAdmin.value)
+const canSeeInternal = computed<boolean>(() => isManager.value || isMember.value || isAdmin.value)
+const showToolbar = computed<boolean>(() => canSeeInternal.value || canEdit.value)
```
bcspragu, line 28: As it currently stands, `showToolbar` is logically equivalent to `canSeeInternal`, but feel free to leave if you expect them to diverge.

frontend/components/initiative/Toolbar.vue
```diff
@@ -0,0 +1,76 @@
...
+</script>
+
+<template>
+  <div
+    v-show="showToolbar"
+    class="p-buttonset initiative-toolbar"
+  >
+    <LinkButton
+      :to="localePath(`/initiative/${initiativeId}`)"
```
bcspragu, line 37: Any reason to not have `LinkButton` just always do `localePath`? It'd cut down on a lot of boilerplate, and I think we already detect internal vs external links.

frontend/components/modal/FakeUsers.vue
```diff
@@ -0,0 +1,99 @@
...
+// TODO(grady) move this to an env-specific location, omit the component if empty.
+const hardcodedLocalUsers = [{
+  id: 'user.c9a959a1c4d6b8be1fc1',
+  email: 'rmi-localtest-a@siliconally.org',
+  password: 'CQdCZd9lq2HOnFB',
+}, {
+  id: 'user.4aed45b19c7ef24e105c',
+  email: 'rmi-localtest-b@siliconally.org',
+  password: 'Dq7CLBX90nFtNk8',
+}]
```
bcspragu, line 20: Do you see any risk with these being public? I think it's probably fine, but you could also `sops` encode credentials and then decrypt them at start-up time. Also, do we need all this fake user infra? What's wrong with just logging out and logging in as other users? It just seems like more pages and more code (and now, more translation) than the limited utility it provides.

frontend/components/modal/FakeUsers.vue
```diff
@@ -0,0 +1,99 @@
...
+
+const { maybeMe } = await getMaybeMe()
+
+const users = computed(() => {
+  const me = maybeMe.value
+  return hardcodedLocalUsers.map((u) => ({
+    itMe: me && u.email.toLowerCase() === me.enteredEmail.toLowerCase(),
```
bcspragu, line 27: `helo, it me`

but also maybe `isLoggedInUser`?

frontend/components/standard/Footer.vue
```diff
@@ -15,14 +16,36 @@ const tt = (s: string) => t(`${prefix}.${s}`)
...
-      />
+    <div class="flex flex-column align-items-center">
+      <div class="flex gap-1">
+        <PVButton
+          v-tooltip.top="tt('Manage Fake Users')"
+          icon="pi pi-user-plus"
+          class="p-1 w-auto"
+          :class="fakeUsersVisible ? '' : 'p-button-text'"
+          @click="() => fakeUsersVisible = !fakeUsersVisible"
+        />
```
bcspragu, line 27: Should this be hidden behind some "only in dev" flag?

frontend/composables/useMSAL.ts
```diff
@@ -19,7 +19,7 @@ import { computed } from 'vue'
 import { type APIKey } from '~/openapi/generated/user'
 
 export const useMSAL = async () => {
-  const isAuthenticated = useState('useMSAL.isAuthenticated', () => false)
+  const { isAuthenticated } = useSession()
```
bcspragu, line 22: This creates a dependency loop:

- `useSession` depends on `usePACTA`
- `usePACTA` depends on `useMSAL`
- `useMSAL` depends on `useSession`

In this case, you should be able to unravel it just by moving `isAuthenticated` back into `useMSAL` (basically just flipping the dependency arrow), but if that doesn't work for some reason, let's talk about it because having loops like this tends to break suddenly and dramatically in hard to debug ways.

frontend/composables/useSession.ts
```diff
@@ -27,7 +25,8 @@ export const useSession = () => {
     // We're the first to request a user, kick of the request and hop in line at the front of the queue.
     return new Promise((resolve) => {
       resolvers.value.push(resolve)
-      void pactaClient.findUserByMe()
+      void usePACTA()
```
bcspragu, line 28: You can't 👏use 👏composables 👏outside 👏of 👏`setup` 👏functions.

If you're trying to avoid the hit of having the whole composable be async, just embrace it. It needs to be async because `usePACTA` needs to be async because `useMSAL` needs to be async because we don't know if we're authenticated (at least on the client) until `useMSAL` initializes, which is an async action.

frontend/envs/env.local
```diff
@@ -8,8 +8,8 @@ MSAL_CLIENT_ID=2d77a4a9-b7be-4451-ad47-c151d8b6c05f
 # See https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md#redirecturi-considerations
 MSAL_REDIRECT_URI=http://localhost:3000/blank
 MSAL_LOGOUT_URI=http://localhost:3000/
-MSAL_MIN_LOG_LEVEL=INFO
+MSAL_MIN_LOG_LEVEL=WARN
```
bcspragu, line 11: I'd consider leaving this as is. It might be noisy, but it's also invaluable when things aren't working as expected.

frontend/lang/en.json
```diff
@@ -21,6 +31,16 @@
         "If Persists": "If this issue persists, please report this issue by",
         "File a Bug": "filing a bug in the PACTA repository"
     },
+    "ModalFakeUsers": {
+        "Heading": "Fake Users",
+        "Subheading": "Quickly switch between profiles for testing.",
+        "Description": "Microsoft Active Directory doesn't allow us to programatically log in as other users, so you will need to manually log in as one of the users below. These are only in the Local environment (for now).",
```
bcspragu, line 37: I haven't actually looked into this at all, how sure are you of this? I know the `msalInstance/PublicClientApplication` doesn't have any obvious `signInWithUserAndPassword` methods, but that doesn't mean there's nothing to be done here.

frontend/lang/en.json
```diff
@@ -33,15 +53,49 @@
...
+        "Yes Invitations": "This initiative requires an invitation (a unique, one-time code) to join. You can create these invitations below.",
+        "You can change": "You can change this behavior on the",
+        "edit initiative page": "edit initiative page",
+        "Create Initiative Invitations": "Create Initiative Invitations",
+        "New Invitations": "New Invitations",
+        "New Invitations Help": "Create new invitations by clicking the button below. You can create as many invitations as you want, but each invitation can only be used once. Invitation codes must only contain alphanumeric characters (a-z, A-Z, 0-9), underscores and dashes. You can create multiple invitations at once by separating them with tabs, commas, newlines, or spaces.",
+        "Create": "Create",
+        "Unused": "Unused",
+        "Includes Illegal Characters": "Includes illegal characters"
+    },
```
bcspragu, line 100: General question about i18n: does the vue/nuxt-i18n stuff allow you to provide context for translations? I can see that being useful for some of these, where without knowing how/where/when we're using it, it might be harder to translate well.

frontend/lib/editor/user.ts
```diff
@@ -0,0 +1,58 @@
+import { type User } from '@/openapi/generated/pacta'
+import { type EditorFieldsFor, Validation, type EditorComputedValues } from './common'
+import { getEditorComputedValues } from './utils'
+
+export type EditorUser = EditorFieldsFor<User>
+
+// TODO(grady) Add i18n for forms.
```
bcspragu, line 7: optional: I feel like this one deserves an honest-to-god GH issue.

frontend/lib/time/index.ts
```diff
@@ -0,0 +1,57 @@
+const defaultLocale = 'en-US'
+
+const getLocale = (): string => {
+  const { locale } = useI18n()
```
bcspragu, line 4: It's generally not a good idea to call composables from anywhere except

1) other composables
2) plugins (and only if you're really careful and know what you're doing
3) setup functions

The risk of course being that this code gets called in an non-setup function, which is doubly likely because this function is then used by a bunch of functions below.

The best solution is likely to just move all of this into it's own `useLocale` or `useTime` composable, which calls `useI18n` at the top level and then returns all these other functions.

frontend/nuxt.config.ts
```diff
@@ -51,10 +52,10 @@ export default defineNuxtConfig({
...
     vueI18n: './i18n.config.ts',
     locales: [
-      { code: 'en', iso: 'en-US', file: { path: 'en.json', cache: false }, flag: '🇬🇧', name: 'English' },
-      { code: 'fr', iso: 'fr-FR', file: { path: 'fr.json', cache: false }, flag: '🇫🇷', name: 'Français' },
-      { code: 'es', iso: 'es-ES', file: { path: 'es.json', cache: false }, flag: '🇩🇪', name: 'Deutsch' },
-      { code: 'de', iso: 'de-DE', file: { path: 'de.json', cache: false }, flag: '🇪🇸', name: 'Español' },
+      { code: 'en', iso: 'en-US', file: { path: 'en.json', cache: false }, name: 'English' },
+      { code: 'fr', iso: 'fr-FR', file: { path: 'fr.json', cache: false }, name: 'Français' },
+      { code: 'es', iso: 'es-ES', file: { path: 'es.json', cache: false }, name: 'Deutsch' },
+      { code: 'de', iso: 'de-DE', file: { path: 'de.json', cache: false }, name: 'Español' },
```
bcspragu, line 58: ¿Sprechen ze ~deutsch~ español?

(which is to say the names are switched)

frontend/pages/initiative/[id]/index.vue
```diff
@@ -0,0 +1,57 @@
...
+const { fromParams } = useURLParams()
+
+const id = presentOrCheckURL(fromParams('id'))
+const prefix = `initiative/${id}/invitations`
+const [
+  { data: initiative },
+] = await Promise.all([
+  useSimpleAsyncData(`${prefix}.getInitiative`, () => pactaClient.findInitiativeById(id)),
+  useSimpleAsyncData(`${prefix}.getInvitations`, () => pactaClient.listInitiativeInvitations(id)),
+  useSimpleAsyncData(`${prefix}.getRelationships`, () => pactaClient.listInitiativeUserRelationshipsByInitiative(id)),
```
Comment (bcspragu) [inline frontend/pages/initiative/[id]/index.vue:14]: Stuff like this confirms my previous grumblings we have the wrong API surface, we shouldn't need three API calls to load a simple page. How about just `GET /initiatives/{id}` returns the initiative, invitations, and relationships, to the extent the caller has access?

frontend/pages/initiative/[id]/internal.vue
```diff
@@ -0,0 +1,20 @@
...
+
+const id = presentOrCheckURL(fromParams('id'))
+const prefix = `initiative/${id}/invitations`
+const [
+  { data: initiative },
+] = await Promise.all([
+  useSimpleAsyncData(`${prefix}.getInitiative`, () => pactaClient.findInitiativeById(id)),
+  useSimpleAsyncData(`${prefix}.getInvitations`, () => pactaClient.listInitiativeInvitations(id)),
+  useSimpleAsyncData(`${prefix}.getRelationships`, () => pactaClient.listInitiativeUserRelationshipsByInitiative(id)),
+])
```
Comment (bcspragu) [inline frontend/pages/initiative/[id]/internal.vue:13]: Any reason this can't just be a hidden/not-hidden section on the `/initiative/[id]` page? Esp since it appears to need the exact same set of data.

frontend/pages/initiative/[id]/invitations.vue
```diff
@@ -0,0 +1,252 @@
+<script setup lang="ts">
+import { humanReadableTimeFromStandardString } from '@/lib/time'
+import type DataTable from 'primevue/datatable'
+// const router = useRouter()
```
Comment (bcspragu) [inline frontend/pages/initiative/[id]/invitations.vue:4]: remove?

bcspragu (approved):
Main takeaways:

- Lots of progress, great work!
- Need to refactor UserID populating middleware (i.e. dump into a `session` package)
- Seems like we should talk about API design and stuff
- No idea why invitations are the way they are
- No composable loops or calling outside of `setup` for general FE stability

(Approved, but maybe let's chat and do one more round of review before this goes in?)

frontend/components/CopyToClipboardButton.vue
```diff
@@ -0,0 +1,36 @@
...
+interface Props {
+  value: string
+  cta: string
+}
+const props = defineProps<Props>()
+
+const prefix = 'CopyToClipboardButton'
+const tt = (key: string) => t(`${prefix}.${key}`)
+
+const statePrefix = `${prefix}[${useStateIDGenerator().id()}]`
```
bcspragu, line 15: I've always thought our state approach (`useStateIDGenerator`) was pretty heavyhanded, turns out we can just use Vue's `ref` when we don't want it shared across instances, e.g. https://github.com/nuxt/nuxt/issues/15110

This doesn't have an SSR component, but generally when we don't want to share state between components, it's because we're recording user interaction stuff (as here), so there's no SSR component anyway.

frontend/components/initiative/Toolbar.vue
bcspragu: ^^

frontend/components/modal/FakeUsers.vue
bcspragu: ^^

frontend/components/standard/Footer.vue
bcspragu: ^^

frontend/components/standard/Nav.vue
```diff
@@ -27,17 +36,20 @@ const menuItems = computed(() => {
...
-  if (showStandardDebug) {
+  if (isAdmin.value) {
     result.push({
       label: tt('Admin'),
       to: localePath('/admin'),
     })
   }
+  if (maybeMe.value) {
+    result.push({
+      label: tt('My Stuff'),
```
bcspragu, line 49: nit: "Stuff" is a bit colloquial for a gov't/bank-facing app, no?

frontend/composables/useSession.ts
```diff
@@ -1,14 +1,13 @@
...
+  const pactaClient = await usePACTA()
 
   const prefix = 'useSession'
-  const signedIn = useState<boolean>(`${prefix}.signedIn`, () => true)
-
   const currentUser = useState<User | undefined>(`${prefix}.currentUser`, () => undefined)
-  const isAdmin = computed(() => currentUser.value && currentUser.value.admin)
-  const isSuperAdmin = computed(() => currentUser.value && currentUser.value.superAdmin)
+  const isAdmin = computed<boolean>(() => !!currentUser.value && currentUser.value.admin)
+  const isSuperAdmin = computed<boolean>(() => !!currentUser.value && currentUser.value.superAdmin)
```
bcspragu, line 10: nit: Do these need the type annotation? Seems like it should be inferred

frontend/composables/useTime.ts
bcspragu, line 1: Excellent, looks great, 10/10 composable

frontend/pages/initiative/[id]/index.vue
Comment (bcspragu) [reply frontend/pages/initiative/[id]/index.vue:13]: ^^

frontend/pages/initiative/[id]/internal.vue
Comment (bcspragu) [reply frontend/pages/initiative/[id]/internal.vue:13]: ^^

frontend/pages/initiative/[id]/relationships.vue
```diff
@@ -0,0 +1,102 @@
...
+
+const { maybeMe, isAdmin } = await getMaybeMe()
+
+const id = presentOrCheckURL(fromParams('id'))
+const prefix = `initiative/${id}/relationships`
+const [
+  { data: relationships },
+] = await Promise.all([
+  useSimpleAsyncData(`${prefix}.getRelationships`, () => pactaClient.listInitiativeUserRelationshipsByInitiative(id)),
+])
```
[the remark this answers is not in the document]
Comment (bcspragu) [reply frontend/pages/initiative/[id]/relationships.vue:22]: (this was resolved, but it's still a `Promise.all` on a single promise)

frontend/plugins/msal.client.ts
```diff
@@ -49,16 +49,16 @@ export default defineNuxtPlugin((_nuxtApp) => {
...
-              console.debug(message)
+              // console.debug(message)
               return
             case LogLevel.Warning:
               console.warn(message)
               return
             case LogLevel.Trace:
-              console.trace(message)
+              // console.trace(message)
               return
```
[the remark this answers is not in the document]
bcspragu: ^^

bcspragu (approved):
Boosted a few previous comments that should be addressed/we should discuss, but otherwise LGTM

frontend/composables/useIsAuthenticated.ts
```diff
@@ -0,0 +1,5 @@
+export const useIsAuthenticated = () => {
+  const prefix = 'useIsAuthenticated'
+  const isAuthenticated = useState<boolean>(`${prefix}.isAuthenticated`, () => true)
```
bcspragu, line 3: Sorry, straggling comment: why does this default to true?

bcspragu (approved):

session/session.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		return WithUserID(c, user.ID), nil
+	}
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx, err := fn(r.Context())
+			if err != nil {
+				// Optionally log errors here when debugging authentication access.
+				// logger.Warn("couldn't authenticate", zap.Error(err))
```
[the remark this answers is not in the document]
gbdubs: FYI We will have non-public endpoints: initiatives are going to be non-public. Does that change your reasoning?

cmd/server/main.go
bcspragu: This one didn't go away, but I suggested a fix elsewhere

session/session.go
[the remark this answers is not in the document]
bcspragu: Discussed offline, there will be public endpoints. That said, our current middleware has no way to poke holes for public endpoints, so anything non-auth'd will fail at the moment

cmd/server/main.go
gbdubs: Done.

cmd/server/main.go
gbdubs: Yep - Done.

cmd/server/pactasrv/initiative_invitation.go
gbdubs: Done

cmd/server/pactasrv/initiative_invitation.go
gbdubs: Done, but through different means, as an external var.

cmd/server/pactasrv/initiative_invitation.go
gbdubs: We won't want that for public access. I think keeping them distinct will make authz much easier.

cmd/server/pactasrv/initiative_user_relationship.go
gbdubs: > If for some reason we're not frequently loading initiatives + their relationships at the same time, could also do /initiative/{id}/relationships, basically like it is now.

Yep! Public access won't load relationships.

I think this actually fits nicely into the "composite key exception" listed [here](https://github.com/stickfigure/blog/wiki/How-to-%28and-how-not-to%29-design-REST-APIs#rule-2-dont-add-unnecessary-path-segments)

cmd/server/pactasrv/initiative_user_relationship.go
gbdubs: Yes, they will be large. There is a public initiative, which anyone will have the ability to join, and may be segmented by region (i.e. can't be handled as a standalone case).

cmd/server/pactasrv/pactasrv.go
gbdubs: Fixed

cmd/server/pactasrv/user.go
gbdubs: Will Remove.

frontend/components/initiative/Toolbar.vue
gbdubs: Done

frontend/components/initiative/Toolbar.vue
gbdubs: I've found with AS it's been easier for us to have sentence-type computeds for frontend visibility gates. It's one extra line of code, but decouples the two from one another by default (i.e. modifying the visibility for one wouldn't require a change to the code for the other).

frontend/components/initiative/Toolbar.vue
gbdubs: Ah yes - LinkButton is the only place we could reliably do this, so I think it's better for us not to do so JUST here. It would build up inconsistent expectations across how i18n should be handled on links. It's easier to spot naked links when we know they should always be directly wrapped by an i18n handler.

frontend/components/modal/FakeUsers.vue
gbdubs: Yeah I agree I don't see any risk here.

I'm confident you will find this valuable - I already have.

frontend/components/modal/FakeUsers.vue
gbdubs: Not sure what you mean here? I don't think we have that utility?

frontend/components/standard/Footer.vue
gbdubs: Certainly at some point!

frontend/composables/useMSAL.ts
gbdubs: Fixed.

frontend/pages/initiative/[id]/relationships.vue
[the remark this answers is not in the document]
Comment (gbdubs) [reply frontend/pages/initiative/[id]/relationships.vue:22]: See above :)

frontend/components/modal/FakeUsers.vue
bcspragu: Lol I'm just noting that you've named it `"it me"`

gbdubs:
Discussed offline - hitting the timebox on this. Smaller, better PRs going forward.

PR #45, opened 2023-10-30
Integrate portfolio processing into runner

bcspragu:
This PR integrates the runner with the PACTA portfolio processing code, integrating with Azure blob storage to read + write portfolios and reports.

It creates a minimal set of `pactasrv` APIs and frontend pages to test out the process of 1) uploading portfolios and 2) running the PACTA processing on them.

I've tested the `process_portfolio` piece against local Docker (i.e. `dockertask`), but I haven't tested the `create_report` flow yet in it's current incarnation. This PR is already large enough that I'll add the relevant async task stuff in a follow-up PR.


cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+	// Azure has 3-4 ways to authenticate as an identity to their APIs (KMS, storage, etc).
+	//   - When running locally, we use the "Environment" approach, which means we provide AZURE_* environment variables that authenticate against a local-only service account.
+	//   - When running in Azure Container Apps Jobs, we use the "ManagedIdentitiy" approach, meaning we pull ambiently from the infrastructure we're running on (via a metadata service).
+	// See https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-credential-types for more info
+	if azClientSecret := os.Getenv("AZURE_CLIENT_SECRET"); azClientSecret != "" {
+		if creds, err = azidentity.NewEnvironmentCredential(nil); err != nil {
+			return fmt.Errorf("failed to load Azure credentials from environment: %w", err)
+		}
+	} else {
+		// We use "ManagedIdentity" instead of just "Default" because the default
```
gbdubs, line 85: Can't imagine this was fun to debug!

cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+		return fmt.Errorf("unknown task type %q", taskType)
+	}
+
+	logger.Info("running PACTA task", zap.String("task_type", string(taskType)))
+
+	if err := taskFn(ctx); err != nil {
+		return fmt.Errorf("error running task: %w", err)
+	}
+
+	logger.Info("ran PACTA task successfully", zap.String("task_type", string(taskType)))
```
gbdubs, line 130: Optional: maybe include some structured logging info here about runtime?

cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+
+	return &task.ProcessPortfolioRequest{
+		AssetIDs: assetIDs,
+	}, nil
+}
+
+func (h *handler) uploadDirectory(ctx context.Context, dirPath, container string) error {
+	base := filepath.Base(dirPath)
+
+	return filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
```
gbdubs, line 167: I'm all for not prematurely optimizing, but IIUC, we know that a given run is likely to operate over a large number of assets (particularly, portfolio group reports). I think we will want to parallelize this - using walk for enumeration of tasks, and then a helper to do `UploadAllBlobs`

cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+	}
+
+	if err := destF.Close(); err != nil {
+		return fmt.Errorf("failed to close dest file: %w", err)
+	}
+
+	return nil
+}
+
+func (h *handler) processPortfolio(ctx context.Context, req *task.ProcessPortfolioRequest) error {
```
gbdubs, line 232: This is very clear - thank you!

cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+	// Load the portfolio from blob storage, place it in /mnt/raw_portfolios, where
```
gbdubs, line 233: s/portfolio/portfolios/(many places)

cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+	for sc.Scan() {
+		line := sc.Text()
+		idx := strings.Index(line, "writing to file: " /* 17 chars */)
+		if idx == -1 {
+			continue
+		}
+		paths = append(paths, strings.TrimSpace(line[idx+17:]))
+	}
+
+	// NOTE: This code could benefit from some concurrency, but I'm opting not to prematurely optimize.
```
gbdubs, line 272: Same thought here as above, maybe less important, since the cardinality is O(report outputs) not O(portfolio inputs).

cmd/runner/main.go
```diff
@@ -53,16 +70,294 @@ func run(args []string) error {
...
+		PortfolioID: pacta.PortfolioID(pID),
+	}, nil
+}
+
+func (h *handler) createReport(ctx context.Context, req *task.CreateReportRequest) error {
+	baseName := string(req.PortfolioID) + ".json"
+
+	// Load the processed portfolio from blob storage, place it in /mnt/
+	// processed_portfolios, where the `create_report.R` script expects it
+	// to be.
```
gbdubs, line 299: I'm probably just missing context that you have - is a portfolio report always generated off of exactly one processed portfolio? I thought it (like above) could be multiple?

cmd/runner/main.go
gbdubs, line 232: One nit: can we name this better? Like "process" is just "do". I don't want to bikeshed on naming, but I also don't want to commit ourselves to very generic names.

gbdubs (commented):
Truly excellent PR. Thank you for writing this, it feels like the PACTA work is really coming together nicely.

cmd/runner/main.go
bcspragu: Added the `task_id` so we can correlate the timestamps on the log more easily

cmd/runner/main.go
bcspragu: Good to know, I'm still going to let it ride and see what actual performance looks like. For example, if processing latency is 10x upload latency, this will be a bit of work and complexity for no additional gain.

An `UploadAllBlobs` function needs to make some decisions about how many goroutines to run in parallel for the upload, which requires some experimenting, so I'll save all that until we know we need it.

cmd/runner/main.go
bcspragu: For what it's worth, I believe "Process Portfolio" is what RMI calls this step, so this is just consistency in naming. I'm open to suggestions though.

cmd/runner/main.go
bcspragu: That's actually a question I had for you about how we want to architect async task running. I was originally thinking that the async runner would get a "portfolio group ID" or whatever, but that means it needs to have database access, to translate from "portfolio group X" to "list of assets in portfolios"

I'd like it to be as dumb as possible because it's a large Docker image with code from N different teams in N different languages, so ideally we could sculpt the interface to be "here are some files in blob store, do stuff to them and write them to this other blob store", which  minimizes the permissions that we need to give the runner.

My question here is what the end-user experience/data model looks like and if that is a reasonable/achievable goal. Are there reasons we _can't_ process portfolios within a portfolio group individually at this stage? Let's chat if it's not a simple answer.

cmd/runner/main.go
bcspragu: Same comment as above, my knowledge of what the actual data steps look like is fuzzy, and I don't know what the discrete units of compute are

cmd/runner/main.go
bcspragu: Great question! I actually don't know, and I expect this code will require a bunch of tweaks once [Alex's stub interface](https://github.com/AlexAxthelm/stub_PACTA_containers_for_app/) gets reconciled with [the actual code interface](https://github.com/RMI-PACTA/workflow.transition.monitor)

PR #48, opened 2023-10-31
Finish adding Event Grid validation, and add local testing tools

bcspragu:
This PR finishes up the work of validating requests from Event Grid, and adds support for running the server with a public-facing endpoint via the `--with_public_endpoint=<subdomain>` flag, which is proxied via [frp](https://github.com/fatedier/frp).

I didn't use our [frpembed](https://github.com/Silicon-Ally/frpembed) library because there's an issue with `quic-go` not supporting Go 1.20 (may be fixed, didn't want to go down the rabbithole)

### DIFF
--- a/azure/azevents/azevents.go
+++ b/azure/azevents/azevents.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"path"
 	"time"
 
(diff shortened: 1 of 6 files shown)


gbdubs (commented):
LG - looking forward to using.

PR #49, opened 2023-11-01
Publish Event Grid event from runner after processing portfolio

bcspragu:
This PR adds support for sending 'processed-portfolio' notifications from the runner, and receiving/parsing them in `pactasrv`.

While Azure requires validating a webhook endpoint when you add it, it _doesn't_ provide validation on each individual webhook request, so the next PR will contain basic shared secret authentication to avoid The Internet At Large from `curl`-ing `/events/processed-portfolio`.


gbdubs (commented):
Looks clear to me. Thank you!

PR #51, opened 2023-11-01
Add webhook auth for Event Grid  webhooks, and local infra

bcspragu:
This PR adds support for a shared secret that gets sent with webhook HTTP requests, so that we can validate they came from Azure.

It also adds local support for testing webhooks, which works something like:
1. Developer starts a local server with `--with_public_endpoint=XYZ`
2. Script checks if `local-webhook-XYZ` already exists
3. If not, asks user if they want to create one
4. If yes, waits for server to start (needed for verification to succeed), then creates the webhook

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -87,6 +87,8 @@ func run(args []string) error {
 		azStorageAccount           = fs.String("secret_azure_storage_account", "", "The storage account to authenticate against for blob operations")
 		azSourcePortfolioContainer = fs.String("secret_azure_source_portfolio_container", "", "The container in the storage account where we write raw portfolios to")
 
+		azEventWebhookSecret = fs.String("secret_azure_webhook_secret", "", "The shared secret required for incoming webhooks")
+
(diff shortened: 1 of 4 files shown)


cmd/server/main.go
```diff
@@ -87,6 +87,8 @@ func run(args []string) error {
 		azStorageAccount           = fs.String("secret_azure_storage_account", "", "The storage account to authenticate against for blob operations")
 		azSourcePortfolioContainer = fs.String("secret_azure_source_portfolio_container", "", "The container in the storage account where we write raw portfolios to")
 
+		azEventWebhookSecret = fs.String("secret_azure_webhook_secret", "", "The shared secret required for incoming webhooks")
```
gbdubs, line 90: IIUC, the structure here will make key rotation hard without failing a few requests in there, since we have exactly one secret we consider valid. Consider bifurcating this into "what auth secret should be set" and "which auth secret**s** should be accepted" to at least allow for the possibility of rotation sans downtime.

bcspragu: An excellent suggestion, I'll send you a follow-up today.

bcspragu: https://github.com/RMI-PACTA/app/pull/53

PR #59, opened 2023-11-15
Standardizes YAML Formatting

gbdubs:
Runs [google/yamlfmt](https://github.com/google/yamlfmt) over our openapi definitions.


bcspragu (approved):
LGTM, any interest in checking for bad formatting as part of CI? I wouldn't spend more than ~10 minutes on it though

gbdubs:
Spent more than 10m. Calling it. I think we can just do this from time to time - drift here is of minimal importance.

PR #61, opened 2023-11-16
Creates Structure for Upload Requests

gbdubs:
The request is two step: (a) create the upload URLs and blobs, then after the assets are uploaded (b) kick off a process for parsing explicitly.

Sending the PR for the interface first - trying to get smaller ones going.

### DIFF
--- a/openapi/pacta.yaml
+++ b/openapi/pacta.yaml
@@ -464,7 +464,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/NewPortfolioAsset'
-  /startPortfolioUpload:
+  /portfolio-upload:
     post:
       summary: Starts the process of uploading one or more portfolio files
       description: Creates one or more new incomplete portfolio uploads, and creates upload URLs for the user to put their blobs into.
@@ -475,15 +475,15 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/StartPortfolioUploadRequest'
+              $ref: '#/components/schemas/StartPortfolioUploadReq'
       responses:
         '200':
           description: The assets can now be uploaded via the given signed URLs.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/StartPortfolioUploadResponse'
-  /completePortfolioUpload:
+                $ref: '#/components/schemas/StartPortfolioUploadResp'
+  /portfolio-upload:complete:
     post:
       summary: Called after uploads of portfolios to cloud storage are complete.
       description: Signals that the upload of the portfolios are complete, and that the server should start parsing them.
@@ -494,14 +494,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/CompletePortfolioUploadRequest'
+              $ref: '#/components/schemas/CompletePortfolioUploadReq'
       responses:
         '200':
           description: The process to initiate the parsing of the uploads has been initiated.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CompletePortfolioUploadResponse'
+                $ref: '#/components/schemas/CompletePortfolioUploadResp'
   /test:parsePortfolio:
     post:
       summary: Test endpoint, triggers a task to parse the portfolio
@@ -876,7 +876,7 @@ components:
         asset_id:
           type: string
           description: A unique identifier for the uploaded asset
-    StartPortfolioUploadRequest:
+    StartPortfolioUploadReq:
       type: object
       required:
         - items
@@ -884,8 +884,8 @@ components:
         items:
           type: array
           items:
-            $ref: '#/components/schemas/StartPortfolioUploadRequestItem'
-    StartPortfolioUploadRequestItem:
+            $ref: '#/components/schemas/StartPortfolioUploadReqItem'
+    StartPortfolioUploadReqItem:
       type: object
       required:
         - file_name
@@ -893,7 +893,7 @@ components:
         file_name:
           type: string
           description: The name of the file, including its extension.
-    StartPortfolioUploadResponse:
+    StartPortfolioUploadResp:
       type: object
       required:
         - items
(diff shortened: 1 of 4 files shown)


openapi/pacta.yaml
```diff
@@ -838,6 +876,66 @@ components:
...
+        upload_url:
+          type: string
+          description: The signed URL where the file should be uploaded to, using PUT semantics.
+        incomplete_upload_id:
+          type: string
+          description: A unique identifier for the uploaded asset
+    CompletePortfolioUpload:
+      type: object
+      required:
+        - incomplete_upload_ids
```
bcspragu, line 924: nit: Confusing name, since they are completed from the user's perspective, they're done uploading. I'd just call them `upload_ids`, it's unambiguous here.

openapi/pacta.yaml
```diff
@@ -838,6 +876,66 @@ components:
...
+        file_name:
+          type: string
+          description: The name of the file, including its extension, used as a round-trip id.
+        upload_url:
+          type: string
+          description: The signed URL where the file should be uploaded to, using PUT semantics.
+        incomplete_upload_id:
+          type: string
+          description: A unique identifier for the uploaded asset
+    CompletePortfolioUpload:
```
bcspragu, line 921:
```suggestion
    CompletePortfolioUploadRequest:
```

openapi/pacta.yaml
```diff
@@ -838,6 +876,66 @@ components:
...
+      type: object
+      required:
+        - incomplete_upload_ids
+      properties:
+        incomplete_upload_ids:
+          type: array
+          items:
+            type: string
+          description: The incomplete uploads that have been successfully uploaded and are ready for parsing.
+    CompletePortfolioUpload:
```
bcspragu, line 931:
```suggestion
    CompletePortfolioUploadResponse:
```

openapi/pacta.yaml
```diff
@@ -838,6 +876,66 @@ components:
...
+          description: A unique identifier for the uploaded asset
+    CompletePortfolioUpload:
+      type: object
+      required:
+        - incomplete_upload_ids
+      properties:
+        incomplete_upload_ids:
+          type: array
+          items:
+            type: string
```
bcspragu, line 929: I'd consider having this be an object, like you did for `StartPortfolioUploadRequest.items`, just because you can't really add fields after this point to each upload_id (like a timestamp, or hash, or whatever) in a backwards compatible way otherwise.

bcspragu (approved):
LGTM, though I'm not sure I understand the API:

1. Users request to upload N items
2. Users get N upload URLs
3. Users upload N files
4. Users say "Hey, I uploaded <these IDs>" - One weakness here is that clients don't need to report the same number N of files here
5. Users get back some ID? And the task starts running?

A more "bulletproof" API would be something like:

1. Create a portfolio (or portfolio group, or process group whatever primitive files get attached to), get an ID back
2. Request to upload files to that portfolio ID
3. Can do step 2 as many times as is useful, can be batch or not batch, doesn't matter
4. Mark files completed after they're uploaded. We could do this behind the scenes (as noted elsewhere, using blob storage events), but we could also just roll this validation step into step 5 below.
5. Request to start the processing

From the web client's perspective, this whole process could be kicked off in one go once they select their files, it just makes the API slightly harder to misuse, intentionally or accidentally.

gbdubs:
> LGTM, though I'm not sure I understand the API:
> 
> 1. Users request to upload N items
> 2. Users get N upload URLs
> 3. Users upload N files
> 4. Users say "Hey, I uploaded " - One weakness here is that clients don't need to report the same number N of files here
> 5. Users get back some ID? And the task starts running?
> 
> A more "bulletproof" API would be something like:
> 
> 1. Create a portfolio (or portfolio group, or process group whatever primitive files get attached to), get an ID back
> 2. Request to upload files to that portfolio ID
> 3. Can do step 2 as many times as is useful, can be batch or not batch, doesn't matter
> 4. Mark files completed after they're uploaded. We could do this behind the scenes (as noted elsewhere, using blob storage events), but we could also just roll this validation step into step 5 below.
> 5. Request to start the processing
> 
> From the web client's perspective, this whole process could be kicked off in one go once they select their files, it just makes the API slightly harder to misuse, intentionally or accidentally.

OK FWIW, I think this is mostly doing what you think it is, just with a few translations.

- an `Incomplete Upload` is a file that has been uploaded, but hasn't parsed successfully, either because it failed parsing, or because it is currently processing, or hasn't yet been processed. This staging ground type of approach allows us to have every `portfolio` in the system correspond to a unit that is fit for any type of processing. The requirements specify things like deleting incomplete uploads by default, and building in that kind of logic over `portfolios` would have been riskier, etc. When an incomplete upload successfully parses into (one or more) portfolios, we delete the incomplete upload.
- The "Marking as completed" is just the request to start processing - it's just #5. There is no #4.
- The reason I think it is stronger to request the set of incomplete uploads to process is because if the incomplete upload fails (i.e. file too big and rejected by cloud storage, ex), the client can be smarter about how it handles partial success cases on the frontend, with no requirement for storage on the backend (i.e. we would have to record more concretely the ties between which blobs were uploaded and which were associated with a non-started task, ex, if we didn't specify which things to run when we say "start the run".

Does that make sense?

openapi/pacta.yaml
gbdubs: Great call.

openapi/pacta.yaml
gbdubs: Great call.

openapi/pacta.yaml
gbdubs: Good Catch.

openapi/pacta.yaml
gbdubs: I'd prefer to keep this as is - I think drift between terminology should be avoided when it's all internal only.

openapi/pacta.yaml
```diff
@@ -464,6 +464,44 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/NewPortfolioAsset'
+  /startPortfolioUpload:
```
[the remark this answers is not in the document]
gbdubs: SGTM.

PR #63, opened 2023-11-17
Refactor MSAL integration

bcspragu:
This PR basically just moves all Azure auth integration into the `$msal` plugin, and makes it not client-only.

The problem with `useMSAL` being an async composable is that it made everything downstream async, which causes utter chaos when it comes to loading things, as now it's all happening multiple times and potentially 'in parallel'.

I think I had originally tried to make `useMSAL` a composable so that we'd only try to load auth on pages that needed it, but things like the NavBar mean we _always_ need that anyway, so the next logical step is to realize that it should just be a plugin, and happen before we do anything else. But since we use auth stuff everywhere, it can't just be a client-side plugin, it needs to run on the server to, and just fail if we try to do auth things during SSR (much like `useMSAL` did before anyway)

This should fix the problems I was seeing, which were manifesting as issues in `FakeUsers` and `StandardNav` because they call a few composables that are kinda at different levels of the dependency hierarchy.

Diagram:

The downside of this approach is mainly that `plugins/msal.ts` is now monstrously large, but all the auth stuff is now contained in one file, which will make it easy to backport to OPGEE.


gbdubs (approved):
Nice - solid refactor. Glad to not be doing `await` on every composable.

PR #66, opened 2023-11-17
Serialize errors we encounter

bcspragu:
This PR makes it so that we serialize errors if we're going to send them across the SSR boundary. The problem this fixes is that some errors (like Nuxt reporting composable usage outside a `setup` function) have functions and other non-POJO/non-serializable info with them, which makes our `useState<Error>` implode, ironically hiding the original error.

This introduces a new non-dev dependency (which I don't do lightly) on `serialize-error`, a small (<1kB min + gzip'd) and popular dependency for doing the surprisingly nontrivial task of serializing different types of errors and preserving type information.

### DIFF
--- a/frontend/app.vue
+++ b/frontend/app.vue
@@ -1,8 +1,10 @@
 <script setup lang="ts">
+import { serializeError } from 'serialize-error'
+
 const { loading: { clearLoading }, error: { errorModalVisible, error } } = useModal()
 
 const handleError = (err: Error) => {
-  error.value = err
+  error.value = serializeError(err)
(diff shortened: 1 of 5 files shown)


gbdubs (approved):
Solid solution, would it be possible to port this to our other projects? I know we run into this intermittently across every single surface

bcspragu:
> would it be possible to port this to our other projects?

Most definitely! I want to let it marinade here a bit to make sure it does what I hope it will do in a variety of use cases (including trying to render specific errors based on type on the client after hydration), but we'll see

PR #69, opened 2023-11-20
Creates Delete Blob Interface

gbdubs:


azure/azblob/azblob.go
```diff
@@ -87,6 +87,30 @@ func (c *Client) ReadBlob(ctx context.Context, uri string) (io.ReadCloser, error
...
+
+func (c *Client) DeleteBlobs(ctx context.Context, uris []string) error {
+	// Implement parallel delete if slow - not prematurely optimizing.
+	for i, uri := range uris {
+		if err := c.DeleteBlob(ctx, uri); err != nil {
+			return fmt.Errorf("deleting blob %d/%d: %w", i, len(uris), err)
+		}
+	}
+	return nil
+}
```
bcspragu, line 112: If it doesn't use anything Azure-specific, it doesn't need to live in the Azure implementation. You could throw this in `blob` or a separate `blobutil`, or just have it be a helper on `pactasrv`, e.g.

```suggestion
type BlobDeleter interface {
	DeleteBlob(ctx context.Context, uri string) error
}

func DeleteBlobs(ctx context.Context, bd BlobDeleter, uris []string) error {
	// Implement parallel delete if slow - not prematurely optimizing.
	for i, uri := range uris {
		if err := bd.DeleteBlob(ctx, uri); err != nil {
			return fmt.Errorf("deleting blob %d/%d: %w", i, len(uris), err)
		}
	}
	return nil
}
```

The rational being that a small, orthogonal interface is better as its less for alternative implementations (test blobs, s3, etc) to implement as well. Plus, any other implementation would have an identical impl in this cal

bcspragu (approved):

PR #70, opened 2023-11-20
Updates Parse Runner Structure

gbdubs:


cmd/runner/main.go
```diff
@@ -250,15 +247,15 @@ func (h *handler) downloadBlob(ctx context.Context, srcURI, destPath string) err
 	return nil
 }
 
+// TODO: Send a notification when parsing fails.
 func (h *handler) parsePortfolio(ctx context.Context, taskID task.ID, req *task.ParsePortfolioRequest) error {
 	// Load the portfolio from blob storage, place it in /mnt/raw_portfolios, where
 	// the `process_portfolios.R` script expects it to be.
-	for _, assetID := range req.AssetIDs {
-		srcURI := blob.Join(h.blob.Scheme(), h.sourcePortfolioContainer, assetID)
```
bcspragu, line 257: If we're not using the `sourcePortfolioContainer` any more, can you remove the flag and associated config + terraform?

But also, I like the previous structure better because it means the runner is only ever pulling from a single bucket, whereas moving it to an opaque blob means the caller can set it to whatever, which introduces a (small) new attack vector. What's the motivation for the opaque `BlobURI`s?

cmd/runner/main.go
```diff
@@ -291,21 +288,38 @@ func (h *handler) parsePortfolio(ctx context.Context, taskID task.ID, req *task.
...
 	}
 
 	events := []publisher.Event{
 		{
 			Data: task.ParsePortfolioResponse{
-				TaskID:   taskID,
-				AssetIDs: req.AssetIDs,
-				Outputs:  out,
+				TaskID:  taskID,
+				Request: req,
```
bcspragu, line 321: Why duplicate the whole request in the response? Wouldn't it be better to just have a standard API interface between srv + runner like anything else, where we input the stuff we need and output the stuff we need?

cmd/runner/main.go
```diff
@@ -324,6 +338,23 @@ func (h *handler) parsePortfolio(ctx context.Context, taskID task.ID, req *task.
 	return nil
 }
 
+func countLines(path string) (int, error) {
```
bcspragu, line 341: depending on what this is used for, we probably want to account for the CSV header row? Only really matters for small N

cmd/runner/taskrunner/taskrunner.go
```diff
@@ -83,18 +83,18 @@ func New(cfg *Config) (*TaskRunner, error) {
 }
 
 func (tr *TaskRunner) ParsePortfolio(ctx context.Context, req *task.ParsePortfolioRequest) (task.ID, task.RunnerID, error) {
-	var buf bytes.Buffer
-	if err := json.NewEncoder(&buf).Encode(req.AssetIDs); err != nil {
-		return "", "", fmt.Errorf("failed to encode asset IDs: %w", err)
+	var taskBuffer bytes.Buffer
```
bcspragu, line 86: `*twitches in Go*`: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

For a short function like this, `buf` is unambiguous and to the point

cmd/runner/taskrunner/taskrunner.go
```diff
@@ -83,18 +83,18 @@ func New(cfg *Config) (*TaskRunner, error) {
...
 	return tr.run(ctx, []task.EnvVar{
 		{
 			Key:   "TASK_TYPE",
 			Value: string(task.ParsePortfolio),
 		},
 		{
-			Key:   "ASSET_IDS",
-			Value: buf.String(),
+			Key:   "PARSE_PORTFOLIO_REQUEST",
+			Value: taskBuffer.String(),
```
bcspragu, line 97: It's fine to dump everything in a single env var as long as all the contents are bounded. The second we have arrays of things, we're headed for trouble, as [there's a 128KB limit on env var size](https://askubuntu.com/a/1385554).

Granted, this was true before, but for anything _really_ large, I'd serialize it to blob store and send over the URI

bcspragu (approved):

cmd/runner/main.go
gbdubs: Round-tripped data in an async request paradigm has huge benefits:

- In this case, it removes the requirement we write down details of the task in our database, since the info is available on the task completing. This removes the need for (a) a DB write and (b) error handling in case the DB state mismatches with the async task state (as it could do in the case of a transaction failure, ex).
- Doesn't require the async processing to know what is going to be important on the return trip (this turned out to be hugely important for a variety of things we built into async CM processing)
- No code changes needed to round trip more data (though you still need a binary deploy)
- With payloads as small as they are right now, there isn't a real cost (if that changes in the future I agree we can reevaluate)

In short this is the easier to maintain, more future proof way of doing it. As you say, premature optimization is at the heart of many issues.

cmd/runner/main.go
gbdubs: good thought - I'll subtract 1. It's only used as a metric we can include in the metadata to help users quickly differentiate portfolios from one another using the most basic metrics.

cmd/runner/taskrunner/taskrunner.go
gbdubs: I care less about this than you do, so changed.

I haven't yet bought into the short var name philosophy of go - it seems like just a different preference.

cmd/runner/taskrunner/taskrunner.go
gbdubs: Agreed generally. I've now added in a harder failure mode to alert us if/when this hits that limit.

cmd/runner/main.go
gbdubs: We could pursue the approach you describe here, but it'd require us fully reworking the blob storage architechure, and doing so in a way that isn't clearly better.

Why?

In order to know how to access or delete a blob we need to know the full URI, including the scheme, the container, and the blob ID. Since blobs are present in more context than just the source portfolio, we couldn't infer or extract that information into the code (without just duplicating the logic of what a blob is and how it is handled). Thus, we either can represent that in the DB as a full string in the DB, or chunk it into parts (perhaps represented by enum values for the portfolio container, ex), so that as you suggest here we would be more certain that we're always pulling from the same container.

That doesn't provide additional security, however, it just adds structure. That structure would make migrations harder, encourage introspection/more manual munging, and be more error prone in situations where they are read, deleted or retrieved.

PR #78, opened 2023-12-07
Frontend for Uploads

gbdubs:
Finishes up the frontend for uploading + parsing files.


frontend/components/incompleteupload/Editor.vue
```diff
@@ -0,0 +1,49 @@
...
+const model = computed({
+  get: () => props.editorIncompleteUpload,
+  set: (editorIncompleteUpload: EditorIncompleteUpload) => { emit('update:editorIncompleteUpload', editorIncompleteUpload) },
+})
+</script>
+
+<template>
+  <div>
+    <FormEditorField
+      help-text="The name of this uploaded source file."
```
bcspragu, line 22: These strings don't seem particularly localizable, what's the plan there?

(also, a more general comment that I'm really happy on the `FormEditorField` paradigm you landed on, looks excellent here and elsewhere)

frontend/lib/editor/incomplete_upload.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export type EditorIncompleteUpload = EditorFieldsFor<IncompleteUpload>
+
+const createEditorIncompleteUpload = (incompleteUpload: IncompleteUpload): EditorIncompleteUpload => {
+  return {
+    id: {
+      name: 'id',
+      label: 'ID',
+      originalValue: incompleteUpload.id,
+      currentValue: incompleteUpload.id,
```
bcspragu, line 13: I thought we had killed all this boilerplate with a helper somewhere? I can understand the need for `name` and `label`, but having:

```
originalValue: blah.field,
currentValue: blah.field,
```

seems error-prone and noisy/verbose. Since you're already specifying the keys (e.g. `id`, `name`, etc), you could iterate over them like:

```
const thing = {
	id: {...},
	name: {...},
	description: {...},
}

// This part would go in some helper
for (const key in thing) {
	thing.originalValue = incompleteUpload[key]
	thing.currentValue = incompleteUpload[key]
}
```

frontend/lib/filesize/index.ts
```diff
@@ -0,0 +1,9 @@
+export const formatFileSize = (bytes: number): string => {
+  if (bytes === 0) return 'Empty'
+  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB']
+
+  const k = 1024
```
bcspragu, line 5: Apologies for the hideous pedantry, especially since you aren't technically wrong here, but I'd use `kB` if you're doing base-10, and `KiB` if you're doing base-2 [1].

When talking about file sizes, one normally uses the base 10 variant, e.g. [this random SSD](https://www.samsung.com/us/computing/memory-storage/solid-state-drives/990-pro-w-heatsink-pcie-4-0-nvme-ssd-4tb-mz-v9p4t0cw/) has 4 trillion bytes

Obligatory xkcd: https://xkcd.com/394/

[1] https://cseducators.stackexchange.com/a/4431

frontend/pages/incomplete-uploads.vue
```diff
@@ -0,0 +1,174 @@
...
+      @click="deleteAll"
+    />
+    <StandardDebug
+      :value="data"
+      label="Editor Objects"
+    />
+  </standardcontent>
+</template>
+
+<style lang="scss">
```
bcspragu, line 166: nit: Any chance we can [use `scoped` here](https://vuejs.org/api/sfc-css-features.html)? It's great for avoiding CSS collisions as an app grows.

frontend/pages/portfolios.vue
```diff
@@ -0,0 +1,170 @@
...
+        icon="pi pi-arrow-left"
+        to="/upload"
+        label="Upload New Portfolios"
+      />
+      <PVButton
+        class="p-button-outlined"
+        label="How To Run a Report"
+        icon="pi pi-question-circle"
+        icon-pos="right"
+      />
```
bcspragu, line 142: I'd add a TODO or something to hook this up to something, just so we don't forget about it

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+    let icon = 'pi pi-spinner pi-spin'
+    if (fileState.status === FileStatus.Done) {
+      icon = 'pi pi-check-circle text-success'
+    } else if (fileState.status === FileStatus.Error) {
+      icon = 'pi pi-times text-error'
+    } else if (fileState.status === FileStatus.Selected) {
+      icon = 'pi pi-circle'
+    }
+    let otherError: string | undefined
+    // TODO(grady) validate this server side too.
```
bcspragu, line 89: I filed #79 if you want to reference that here, because I wanted to add some context for how we have to do this when working with Azure.

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+  errorMessage.value = ''
+  isProcessing.value = true
+  startedProcessing.value = true
+  setAllStates(FileStatus.Waiting)
+  const startPortfolioUploadResp = await pactaClient.startPortfolioUpload({
+    items: fileStates.value.map((fileState: FileState) => ({
+      file_name: fileState.file.name,
+      // TODO(grady) consider adding file size here as a validation step.
+      // TODO(grady) consider adding file type validation here.
+      // TODO(grady) consider adding duplicate file checking here.
```
bcspragu, line 161: Isn't that happening above, or is that at a different point in the upload lifecycle?

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+  isProcessing.value = false
+}
+
+const doParsing = async () => {
+  fileStates.value.forEach((_, i) => {
+    fileStates.value[i].status = FileStatus.Validating
+  })
+  await pactaClient.completePortfolioUpload({
+    items: fileStates.value.map((fileState) => ({
+      incomplete_upload_id: presentOrFileBug(fileState.incompleteUploadId),
```
bcspragu, line 236: How do the `presentOrX` modal helpers function in a loop? Just want to make sure we don't just start savagely flashing the user's screen if 10 uploads fail or something.

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+
+const waitForValidationToCompleteOrTimeout = async () => {
+  const timeout = 1000 * 60 * 5
+  const start = Date.now()
+  do {
+    await refreshStateFromIncompleteUploads()
+    if (Date.now() - start > timeout) {
+      throw new Error('Timeout waiting for uploads to complete')
+    }
+    await new Promise((resolve) => setTimeout(resolve, 1000))
```
bcspragu, line 251: [no action required]

I've never seen `await ... setTimeout` essentially used like `sleep` and it makes me feel weird, even if its totally reasonable.

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+  } while (fileStates.value.some((fileState) => fileState.status === FileStatus.Validating))
+  isProcessing.value = false
+}
+
+const refreshStateFromIncompleteUploads = async () => {
+  const resp = await pactaClient.listIncompleteUploads()
+  console.log('resp', resp)
```
bcspragu, line 258: remove

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+    const idx = fileStates.value.findIndex((fileState) => fileState.incompleteUploadId === incompleteUpload.id)
+    // Note - this item might not be in the list if the user hasn't cleaned up prior incomplete portfolios.
+    if (idx >= 0) {
+      if (incompleteUpload.failureCode) {
+        fileStates.value[idx].status = FileStatus.Error
+        fileStates.value[idx].errorMessage = incompleteUpload.failureMessage
+      } else if (incompleteUpload.completedAt) {
+        fileStates.value[idx].status = FileStatus.CleanUp
+      }
+    }
```
bcspragu, line 270: optional nit: handle error conditions early and unindent main body, e.g.

```suggestion
    // Note - this item might not be in the list if the user hasn't cleaned up prior incomplete portfolios.
  if (idx < 0) {
    continue
  }
  if (incompleteUpload.failureCode) {
    fileStates.value[idx].status = FileStatus.Error
    fileStates.value[idx].errorMessage = incompleteUpload.failureMessage
  } else if (incompleteUpload.completedAt) {
    fileStates.value[idx].status = FileStatus.CleanUp
  }
}
```

frontend/pages/upload.vue
```diff
@@ -0,0 +1,407 @@
...
+  }
+}
+</script>
+
+<template>
+  <StandardContent>
+    <TitleBar title="Upload Portfolios" />
+    <p>
+      This is a page where you can upload portfolios to test out the PACTA platform.
+      This Copy will need work, and will need to link to the documentation.
```
bcspragu, line 291: Maybe have some big root GH issue we link to for all the places we'll need a real person to look at the copy for?

bcspragu (approved):
LGTM!

frontend/lib/filesize/index.ts
gbdubs: Not Pedantry! Sounds correct :)

frontend/pages/upload.vue
gbdubs: Good call - fixed that one!

frontend/pages/upload.vue
gbdubs: Since they throw, I assume they break the cycle of execution, and are caught at the higher level, which doesn't resume the execution of the thing that threw, just clears the error. Standard throw semantics no?

frontend/pages/upload.vue
gbdubs: Nice! This is my (and it looks like the internet's) One liner:

https://stackoverflow.com/questions/37764665/how-to-implement-sleep-function-in-typescript

frontend/pages/upload.vue
gbdubs: Great call.

frontend/pages/upload.vue
gbdubs: Great call

frontend/pages/upload.vue
gbdubs: Great idea. Filed #80

frontend/lib/editor/incomplete_upload.ts
gbdubs: As soon as this PR is submitted I'm going to work on #57, which I think should provide the best opportunity for making this possible - I'll try to address this concern in that PR.

frontend/pages/upload.vue
bcspragu: Good call, I forgot they throw

PR #81, opened 2023-12-09
Form i18n, Translation Tooling

### LINKED ISSUE #57: Create i18n for forms

gbdubs:
### Form i18n

- Resolves #57 , by creating a cleaner paradigm for form i18n
- As part of this, splits a form field into `value` (pure v-modeled) and `field` (configuration part, computed), to avoid side effects in one from impacting the other (i.e. switching locale messing with form value, or form value resetting re-doing `i18n` calls). This yields a ton of diffs, but is actually pretty minimal.
- Addresses the unresolved comment from #78 , removing some form boilerplate.

### Translation Tooling

- Adds a custom handler (using `inject/provide` - open to better ideas, but because it's used in a module, and installed/made available through a plugin, I couldn't get any other configuration working). The handler adds missing messages to a `useState` data structure.
- Creates a modal that synthesizes the missing messages and the existing messages into an idealized state (which is, critically, sorted definitively for easier diffing)
- Renames all translation prefixes to match the path of the file they're in, to make editing them more straightforward, and encourage consistent naming.


frontend/components/DownloadButton.vue
```diff
@@ -0,0 +1,42 @@
...
+
+const prefix = 'components/DownloadButton'
+const tt = (key: string) => t(`${prefix}.${key}`)
+
+const statePrefix = `${prefix}[${useStateIDGenerator().id()}]`
+const downloaded = useState<boolean>(`${statePrefix}.downloaded`, () => false)
+const message = computed(() => downloaded.value ? tt('Downloaded') : props.cta)
+const icon = computed(() => downloaded.value ? 'pi pi-check' : 'pi pi-download')
+
+const download = async () => {
```
bcspragu, line 21: nit: doesn't need to be async, there's [a lint for that](https://github.com/typescript-eslint/typescript-eslint/blob/HEAD/packages/eslint-plugin/docs/rules/require-await.md), which we should probably turn on.

frontend/components/form/EditorField.vue
```diff
@@ -1,42 +1,40 @@
...
+import { type EditorField, type EditorValue, isValid } from '@/lib/editor'
+
+const { t } = useI18n()
+const tt = (key: string) => t(`components/form/EditorField.${key}`)
 
 interface Props {
   editorField: EditorField<any, keyof any>
-  helpText?: string
-  startHelpTextExpanded?: boolean
+  editorValue: EditorValue<any, keyof any>
```
bcspragu, line 10: I'm no TypeScript whiz, but I'd imagine `keyof any` would just evaluate to `any`. If there's supposed to be some relationship between these two generic parameters, I'd capture it with something like `<T, keyof T>`.

frontend/components/incompleteupload/Editor.vue
```diff
@@ -1,48 +1,58 @@
...
-  (e: 'update:editorIncompleteUpload', ei: EditorIncompleteUpload): void
+  (e: 'update:editorValues', evs: EditorValues): void
 }
 const props = defineProps<Props>()
 const emit = defineEmits<Emits>()
 
-const model = computed({
-  get: () => props.editorIncompleteUpload,
-  set: (editorIncompleteUpload: EditorIncompleteUpload) => { emit('update:editorIncompleteUpload', editorIncompleteUpload) },
+const efs = computed(() => props.editorFields)
```
bcspragu, line 22: Why wrap this instead of just using `props.editorFields` directly? To make it read-only? And elsewhere

frontend/components/modal/MissingTranslations.vue
```diff
@@ -0,0 +1,141 @@
...
+const { $axios, $missingTranslations } = useNuxtApp()
+const { missingTranslations: { missingTranslationsVisible } } = useModal()
+const { t } = useI18n()
+
+const prefix = 'components/modal/MissingTranslations'
+const existing = useState<Map<string, string>>(`${prefix}.existing`, () => new Map())
+const languages = ['en', 'es', 'fr', 'de']
+const tt = (key: string) => t(`${prefix}.${key}`)
+
+onMounted(async () => {
```
bcspragu, line 11: FWIW, you could make this work on SSR too, likely with [useFetch](https://nuxt.com/docs/api/composables/use-fetch) (though you could also specify a full path, but that's grosser)

frontend/components/modal/MissingTranslations.vue
```diff
@@ -0,0 +1,141 @@
...
+  for (const lang of languages) {
+    const langData = await $axios({
+      method: 'get',
+      url: `/_nuxt/lang/${lang}.json`,
+    }).then((r) => {
+      return JSON.stringify(r.data)
```
bcspragu, line 17: nit: Looks like the data is being parsed into JSON just to be turned back to a string again (and then, turned back into JSON below!), which is wasteful (esp for larger responses). You can [disable Axios's JSON parsing here](https://stackoverflow.com/a/45548588), or work with the structured data, or `useFetch` might also solve this.

Though since you're later re-parsing this into JSON, I'd just leave it as JSON here.

frontend/i18n.config.ts
```diff
@@ -2,4 +2,12 @@ export default defineI18nConfig(() => ({
   locale: 'en',
   fallbackLocale: 'en',
   objectNotation: true,
+  missing: (locale, key, vm) => {
+    // TODO(grady) figure out how to skip this if we're in production + just log.
```
bcspragu, line 6: If `i18n.config.ts` is loaded at runtime (which I'm guessing it is, based on this), you should be able to use something like `process.env.NODE_ENV == 'prod'` or something, and then set those in our `package.json` scripts

frontend/plugins/handle-missing-translation.ts
```diff
@@ -0,0 +1,32 @@
...
+    presentOrFileBug(missingTranslations.value.get(locale)).add(key)
+    if (presentOrFileBug(missingTranslations.value.get(locale)).size > size) {
+      numberMissing.value++
+      console.log(`missing ${locale} translation for "${key}" (${numberMissing.value} total)`)
+    }
+  }
+
+  nuxtApp.vueApp.provide('handleMissingTranslation', handleMissingTranslation)
+  const values = computed(() => {
+    console.log('missingTranslations computed')
```
bcspragu, line 21: remove?

bcspragu (approved):
LGTM, and great refactoring for the form editor stuff, even if the sprinkling of `any`s scares me a bit

frontend/components/DownloadButton.vue
gbdubs: GC - turned on the lint.

frontend/components/incompleteupload/Editor.vue
gbdubs: Exactly - readonly :) Added a comment for that.

gbdubs: Just for readability here and elsewhere - found it easier to grok and copy/paste code.

frontend/components/form/EditorField.vue
gbdubs: Great call - done.

frontend/components/modal/MissingTranslations.vue
gbdubs: Good thought - this will be removed in prod, so there isn't an actual production impact, but I'll keep it in mind!

frontend/components/modal/MissingTranslations.vue
gbdubs: Great point - thanks for the pointer. Fixed.

frontend/i18n.config.ts
gbdubs: Good suggestion - done.

frontend/plugins/handle-missing-translation.ts
gbdubs: GC

PR #83, opened 2023-12-18
Backend for Portfolio Groups

gbdubs:
This PR is the backend half of one I worked on this weekend to build out our portfolio group management capabilities.


cmd/server/pactasrv/populate.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/RMI/pacta/oapierr"
+	"github.com/RMI/pacta/pacta"
+	"go.uber.org/zap"
+)
+
+func populateAll[Source any, TargetID ~string, Target any](
```
bcspragu, line 12: This is a clever use of generics, but it's also hard to understand what it's doing, both here and where its called. I'd add a comment or similar.

Also, it looks like we're doing that "populate a pre-existing pointer" thing, which I begrudgingly accepted was useful in a GraphQL system for populating arbitrary recursive responses, but in a standard RESTful server, I can't see the motivation for such gymnastics when populating a static, non-recursive structure.

cmd/server/pactasrv/populate.go
```diff
@@ -0,0 +1,104 @@
...
+			} else {
+				return fmt.Errorf("can't find populated target %s", id)
+			}
+		}
+	}
+	return nil
+}
+
+func populatePortfoliosInPortfolioGroups(
+	s *Server,
```
bcspragu, line 59: Since this function isn't generic, you should be able to define it on the server (i.e. `func (s *Server) populatePortfoliosInPortfolioGroups(...)`) instead of having to pass it in. I'm guessing this is the remnant of some refactoring.

And same elsewhere.

db/sqldb/portfolio.go
```diff
@@ -3,45 +3,50 @@ package sqldb
...
+		portfolio.created_at,
+		portfolio.holdings_date,
+		portfolio.blob_id,
+		portfolio.admin_debug_enabled,
+		portfolio.number_of_rows,
+		portfolio_group_membership.portfolio_group_id,
+		portfolio_group_membership.created_at
+	FROM portfolio
+	LEFT JOIN portfolio_group_membership 
+	ON portfolio_group_membership.portfolio_id = portfolio.id
```
bcspragu, line 29: Noted below, but the downside of this approach is that we repeatedly load all the portfolio information, once for each group membership. As long as the portfolio metadata is a small known quantity (e.g. no giant blob columns), and we don't expect the cardinality of groups per portfolio to be in the thousands, this is likely fine, but otherwise I'd just batch load membership info in a separate query given all the portfolio IDs.

Same deal for `PortfolioGroup`s

db/sqldb/portfolio.go
```diff
@@ -3,45 +3,50 @@ package sqldb
...
+	rows, err := d.query(tx, portfolioQueryStanza+` WHERE id = $1;`, id)
 	if err != nil {
 		return nil, fmt.Errorf("querying portfolio: %w", err)
 	}
 	pvs, err := rowsToPortfolios(rows)
 	if err != nil {
 		return nil, fmt.Errorf("translating rows to portfolios: %w", err)
 	}
-	return exactlyOne("portfolio", id, pvs)
+	return exactlyOne("portfolio", id, valuesFromMap(pvs))
```
bcspragu, line 41: I had initially complained that it seemed silly to query by primary key with `query` instead of `queryRow`, only to then verify we got a single entry. Now I'm doubly complaining because we're loading what we know is zero or one result(s) into a map, to load it into an array, to verify we got a single entry. Not that this is anywhere near a bottleneck, but it would be simpler and more efficient just to use queryRow and check for ErrNoRows or whatever it is.

db/sqldb/portfolio.go
```diff
@@ -69,7 +72,9 @@ func (d *DB) PortfoliosByOwner(tx db.Tx, ownerID pacta.OwnerID) ([]*pacta.Portfo
 	if err != nil {
 		return nil, fmt.Errorf("translating rows to portfolios: %w", err)
 	}
-	return pvs, nil
+	// Note the map interface here is ~required in the deserialization process to track multiple membersihps,
+	// so we're not just converting to a map and back.
```
bcspragu, line 76: I appreciate getting out in front of the inevitable review comment, though instead of having the SQL query repeatedly returns the whole portfolio alongside each group it is in, you could split it into two queries, one for the portfolios, one for the memberships. But that approach still has a mapping step, so your point stands, LGTM

Also niiiit on `s/membersihps/memberships`

db/sqldb/portfolio.go
```diff
@@ -153,32 +158,65 @@ func (d *DB) DeletePortfolio(tx db.Tx, id pacta.PortfolioID) ([]pacta.BlobURI, e
...
+		&p.Portfolio.Name,
+		&p.Portfolio.Description,
+		&p.Portfolio.CreatedAt,
 		&hd,
-		&p.Blob.ID,
-		&p.AdminDebugEnabled,
-		&p.NumberOfRows,
+		&p.Portfolio.Blob.ID,
+		&p.Portfolio.AdminDebugEnabled,
+		&p.Portfolio.NumberOfRows,
```
bcspragu, line 181: optional: You should be able to implement a custom `Scan` method such that you can just `row.Scan(&p.Portfolio, &mid, &mca)` and it will work. Mostly just syntactic sugar, but cleans things up a bit and allows scanning directly into a Portfolio from multiple places.

Same for PortfolioGroup

openapi/pacta.yaml
bcspragu, line 1: super nit: there's a bunch of trailing whitespace in this file

cmd/server/pactasrv/populate.go
gbdubs: GC on commenting this to make it clearer. 

Not sure of the distinction you're drawing on that one: we have recursive structures in our data model here too: `Portfolio.MemberOf` is a `PortfolioGroupMembership`,which has a `Portfolio` field (same with `PortfolioGroup`). The value of these selsective population things is exactly that we can build a DB layer that is "dumb", and then populate the data we want selectively based on the context.

This is the ~the same pattern we used for Abound, ex, because the data model is actually really similar. Portfolio:Art, Gallery:PortfolioGroup, ex.

cmd/server/pactasrv/populate.go
gbdubs: Good catch!

db/sqldb/portfolio.go
gbdubs: I don't think that's the case - the `portfolio_group_membership` table is just the join table, not the raw `portfolio` table, so even if we stored a blob on the `portfolio` table, this wouldn't pull that in.

db/sqldb/portfolio.go
gbdubs: I agree with the diagnosis of layercaking here, but the benefit IMO outweighs the extra layer. One code pathway for decoding rows allows us to build code that is (a) more resilient to changes (i.e. we couldn't read a field in one case but not in another), (b) less work to maintain (i.e. a change to the data model needs to be made in fewer places)

db/sqldb/portfolio.go
gbdubs: Yay! GC on the typo.

db/sqldb/portfolio.go
gbdubs: Oh interesting, I'd not thought of doing that. I'll file a bug to take a look at that. I'd like to have that in the quiver.

gbdubs: #84

openapi/pacta.yaml
gbdubs: odd. Sorry!

bcspragu (approved):

db/sqldb/portfolio.go
bcspragu: An example to better illustrate my point:

```sql
CREATE TABLE portfolio (
	id TEXT NOT NULL PRIMARY KEY,
	some_giant_blob TEXT NOT NULL
);

CREATE TABLE some_join_table (
	id TEXT NOT NULL PRIMARY KEY,
	portfolio_id TEXT NOT NULL REFERENCES portfolio (id)
);

INSERT INTO portfolio (id, some_giant_blob) VALUES ('p1', 'imagine this is huge');
INSERT INTO some_join_table (id, portfolio_id) VALUES ('sjt1', 'p1'), ('sjt2', 'p1'), ('sjt3', 'p1');

SELECT
	portfolio.id,
	portfolio.some_giant_blob,
	some_join_table.id
FROM portfolio
LEFT JOIN some_join_table
	ON some_join_table.portfolio_id = portfolio.id;

 id |   some_giant_blob    |  id
----+----------------------+------
 p1 | imagine this is huge | sjt1
 p1 | imagine this is huge | sjt2
 p1 | imagine this is huge | sjt3
(3 rows)
```

See how we load the portfolio once for each relationship in the other table? That's potentially a lot of extra data to transmit and parse if the portfolio is large or there are a lot of groups.

gbdubs: Sorry I misunderstood this - I see the point and I know a fix! I'll pursue this afternoon.

PR #85, opened 2023-12-19
Frontend for Portfolio Groups

gbdubs:
image

Notes:

- Fixes a bunch of small i18n bugs
- Directed URL parameterization, so that we can do good rich linking
- Handles recursive JSON data structures well in `StandardDebug`
- Editor + Library for Portfolio Groups
- Sorting + Filtering tables more effectively
- Bulk tools for modifying portfolio group memberships


frontend/components/modal/MissingTranslations.vue
```diff
@@ -1,33 +1,39 @@
...
 const existing = useState<Map<string, string>>(`${prefix}.existing`, () => new Map())
 const languages = ['en', 'es', 'fr', 'de']
 const tt = (key: string) => t(`${prefix}.${key}`)
 
-onMounted(async () => {
-  for (const lang of languages) {
-    const langData = await $axios({
+const onOpen = async () => {
+  await withLoading(async () => {
+    await Promise.all(languages.map((lang) => $axios({
```
bcspragu, line 13: nit: Looks like there's still some unnecessary async/await here. You can just return the promise directly, e.g.

Same elsewhere, like `pages/portfolios.vue`

```suggestion
  await withLoading(() => Promise.all(languages.map((lang) => $axios({
```

frontend/components/portfolio/ListView.vue
```diff
@@ -0,0 +1,224 @@
...
+    return editorObjects.value.filter((editorObject) => ids.includes(editorObject.id))
+  },
+  set: (value: EditorObject[]) => {
+    const ids = value.map((row) => row.id)
+    ids.sort()
+    selectedPortfolioIDs.value = ids
+  },
+})
+
+const editorObjects = computed<EditorObject[]>(() => props.portfolios.map((item) => ({ ...portfolioEditor(item, i18n), id: item.id })))
```
bcspragu, line 54: I'm not sure if this construct is guaranteed to work reliably, since `portfolioEditor` is calling `computed` itself, which will be invoked after `setup` runs. I think `computed` is one of those functions that you're only supposed to run in the `setup` function itself.

As for a fix, the `props` should already be reactive, so just calling `portfolioEditor` for each portfolio in `props.portfolio` at `setup` time _might_ be enough to make it work. Otherwise we might need to shake the paradigm up a bit.

frontend/components/portfolio/group/membership/MenuButton.vue
```diff
@@ -0,0 +1,175 @@
...
+const toggleMenu = (event: Event) => {
+  presentOrFileBug(overlayPanel.value).toggle(event)
+  visible.value = !visible.value
+}
+const hideMenu = () => {
+  visible.value = false
+}
+const showMenu = () => {
+  visible.value = true
+}
```
bcspragu, line 36: nit: Normally I'm advocating for the opposite direction, but these are simple and straightforward enough that they could be inlined, since they're only used in one place.

frontend/components/portfolio/group/membership/MenuButton.vue
```diff
@@ -0,0 +1,175 @@
...
+      await Promise.all(portfolioIds.map((portfolioId) => pactaClient.createPortfolioGroupMembership({ portfolioId, portfolioGroupId })))
+    } else {
+      await Promise.all(portfolioIds.map((portfolioId) => pactaClient.deletePortfolioGroupMembership({ portfolioId, portfolioGroupId })))
+    }
+    emit('changed-memberships')
+    toast.add({
+      severity: 'success',
+      group: add ? 'portfolio-group-membership-added' : 'portfolio-group-membership-removed',
+    })
+    // toggleMenu(event)
```
bcspragu, line 51: remove or uncomment?

frontend/components/portfolio/group/membership/MenuButton.vue
```diff
@@ -0,0 +1,175 @@
...
+    return {
+      id: pg.id,
+      label: pg.name,
+      icon,
+      cmd: changeMemberships(pg.id, addIfClick),
+      hoverText,
+      created: pg.createdAt,
+    }
+  })
+  result.sort((a, b) => a.created < b.created ? 1 : -1)
```
bcspragu, line 88: Not sure what the type of `created` is, but one trick for sorting things is to do `a.created - b.created` or vice versa. I think the cardinality helps it sort more efficiently too.

frontend/components/portfolio/group/membership/MenuButton.vue
```diff
@@ -0,0 +1,175 @@
...
+          <PVButton
+            v-tooltip="option.hoverText"
+            class="text-left p-button-text w-full"
+            @click="option.cmd"
+          >
+            <div class="flex justify-content-start align-items-center gap-3">
+              <div
+                class="flex-0 border-2 border-round flex justify-content-center align-items-center"
+                :class="option.icon === 'empty' ? 'bg-white' : 'bg-primary-500 text-white border-primary-500'"
+                style="height: 1.25rem; width: 1.25rem;"
```
bcspragu, line 131: lol `h-1rem` and `w-1rem` weren't doing it for you?

But more seriously, if these values are useful elsewhere (e.g. for consistency), consider putting them in some utility class or variables in the top level SCSS/CSS.

frontend/components/portfolio/group/membership/MenuButton.vue
```diff
@@ -0,0 +1,175 @@
...
+              >
+                <i
+                  v-if="option.icon === 'full'"
+                  class="pi pi-check"
+                  style="font-size: 1rem"
```
bcspragu, line 136: The Primeflex class `text-base` is 1rem - https://primeflex.org/fontsize

frontend/components/portfolio/group/membership/MenuButton.vue
```diff
@@ -0,0 +1,175 @@
...
+                />
+                <i
+                  v-if="option.icon === 'partial'"
+                  class="pi pi-minus"
+                  style="font-size: .8rem"
```
bcspragu, line 141: The Primeflex class `text-sm` is 0.875rem, FWIW

bcspragu (approved):
Excellent work, really happy to see PACTA coming together so quickly

frontend/components/modal/MissingTranslations.vue
gbdubs: Good catch!

frontend/components/portfolio/group/membership/MenuButton.vue
gbdubs: SG

frontend/components/portfolio/group/membership/MenuButton.vue
gbdubs: It's an ISO string, so it's lexographically accurate, which is great, and we don't need to parse it to do the comparison. I'll comment something to that effect.

frontend/components/portfolio/group/membership/MenuButton.vue
gbdubs: SG - I'll put them elsewhere.

Yeah, my goal was to match the size and vibe of the checkboxes that PV uses, but since we include a third state (dash) for mixed uses, PV didn't have a solution, so I built this.

frontend/components/portfolio/group/membership/MenuButton.vue
gbdubs: GC - I was fine-tuning this and found that 1rem was perfect lol.

frontend/components/portfolio/group/membership/MenuButton.vue
gbdubs: This one I'll skip - getting these to look reasonable was sneaky hard.

frontend/components/portfolio/ListView.vue
gbdubs: Ah, but we can't do that either because

`Getting a value from the props in root scope of setup() will cause the value to lose reactivity`

This does work FWIW. I think I can explain how too: when we need to access the structure (i.e. how many elements in the array), the first `.value` call will cause that to be reactive. When we need to access the content (i.e. a specific element of the array's internals) we're doing an additional `.value` call. That can just be thought of as two computed values, so while I don't think we'd get cross-ref reactivity (i.e. changing a value of an inner ref wouldn't trigger reactivity in an outer ref), that's actually OK so long as we're de-referencing both in any instance that requires access to both values (which we are). See line 112 for an example. The `editorObjects` value in the PDataTable makes the rows reactive, and the additional dereference to `editorValues.value` within the template makes the row value reactive. 

I agree this is more complicated than is ideal, and I'm open to restructuring ideas, I just can't think of how to do that in a way that allows for pure-read `computed`s and bidirectional `computed`s in the same object. And FWIW, this works without error logs, which is nice.

PR #87, opened 2023-12-19
Uses `ARRAY_AGG` to make DB lookups more efficent

### LINKED ISSUE #86: Use Array Agg to avoid multi-pull of all join fields
https://github.com/RMI-PACTA/app/pull/83#discussion_r1430769430

gbdubs:
Fixes #86


bcspragu (approved):
Excellent! This is exactly what I had in mind, looks great, very direct

PR #90, opened 2023-12-19
Basic UGC Limits + Helpers

### LINKED ISSUE #71: Implement Limits on all UGC

gbdubs:
Fixes #71


cmd/server/pactasrv/pacta_version.go
```diff
@@ -47,6 +47,13 @@ func (s *Server) ListPactaVersions(ctx context.Context, request api.ListPactaVer
 // Creates a PACTA version
 // (POST /pacta-versions)
 func (s *Server) CreatePactaVersion(ctx context.Context, request api.CreatePactaVersionRequestObject) (api.CreatePactaVersionResponseObject, error) {
+	if err := anyError(
+		checkStringLimitSmall("name", request.Body.Name),
+		checkStringLimitMedium("digest", request.Body.Digest),
```
bcspragu, line 52: If the digest is a SHA256 value, it'll be a fixed size, which is much smaller than 1000 ASCII characters/bytes, meaning this could be `Small`, and same elsewhere

bcspragu (approved):
LGTM, though did you look at [the OpenAPI codegen tool](https://github.com/deepmap/oapi-codegen) we use at all? It has lots of functionality built-in, I'm wondering if this could be done at the `pacta.yaml` layer, which would be a better option along every axis (no boilerplate, one source of truth, spec generates the code, etc)

cmd/server/pactasrv/pacta_version.go
gbdubs: Great call. Done.

gbdubs:
Discussed offline - OpenAPI spec does support a limitation on string length, but the codegen tools we use do not support it :(
https://swagger.io/docs/specification/data-models/data-types/#:~:text=multipleOf%3A%20%2D5.-,Strings,-A%20string%20of

PR #91, opened 2023-12-23
Front-end for Initiative Portfolio Participation

gbdubs:
- Creates a frontend for initiative memberships, including adding and removing portfolio participation in an initiative.
- Creates a frontend for incomplete upload management, flagging to a user using a yellow banner if there are incomplete uploads for them to delete.
- Renames the 'portfolios' page into the 'my-data' page, and adjusts URLs accordingly.
- Changes the URL param structure to accomodate more range of tab configurations (missing tabs, optional tabs, named tabs)
- Centralizes PVToast so that we don't have them overlap.
- Makes a ton of semantic tweaks to the data management page, using icons + colors more consistently.
- Fixes a number of i18n misses.

Additionally, fixes a few backend errors that arose during testing
- Fixed a cardinality bug where we were imprecisely using `ARRAY_AGG`, which led to strange outcomes when we had aggregated arrays of differing lengths.
- Adds a uniqueness constraint to the portfolio initiative membership table, which was missing.
- Adds better handling in the `conv` layer for partially populated entities.


db/sqldb/portfolio.go
```diff
@@ -22,18 +25,32 @@ func portfolioQueryStanza(where string) string {
...
+		SELECT
+			portfolio_id,
+			ARRAY_AGG(initiative_id) as initiative_ids,
+			ARRAY_AGG(added_by_user_id) as initiative_added_by_user_ids,
+			ARRAY_AGG(created_at) as initiative_created_ats
+		FROM portfolio_initiative_membership
+		WHERE portfolio_id IN (SELECT id FROM selected_portfolio_ids)
+		GROUP BY portfolio_id
+	) itvs ON itvs.portfolio_id = portfolio.id
+	%[1]s;`, where)
```
bcspragu, line 53: I'm not sure if I understand these changes, specifically why these `LEFT JOIN`s are now structured as subqueries, since they just join on the portfolio ID anyway. I know you mentioned a "cardinality bug", but I don't follow how this fixes that, when the subqueries are grouping by the same thing we were initially grouping by anyway (portfolio ID).

db/sqldb/portfolio.go
```diff
@@ -215,16 +232,22 @@ func rowToPortfolio(row rowScanner) (*pacta.Portfolio, error) {
...
+			return nil, fmt.Errorf("initiative ids must be non-null")
 		}
-		if !groupsCreatedAts[i].Valid {
-			return nil, fmt.Errorf("portfolio group membership createdAt must be non-null")
+		if !initiativesCreatedAts[i].Valid {
+			return nil, fmt.Errorf("initiative createdAt must be non-null")
+		}
+		id := pacta.InitiativeID(initiativesIDs[i].String)
+		if seenInitiativeIDs[id] {
+			continue // Skip duplicates, see MULTI_LEFT_JOIN note.
```
bcspragu, line 248: I don't see `MULTI_LEFT_JOIN` occur anywhere else in the codebase.

db/sqldb/portfolio_group.go
```diff
@@ -138,12 +138,15 @@ func rowToPortfolioGroup(row rowScanner) (*pacta.PortfolioGroup, error) {
 	if err != nil {
 		return nil, fmt.Errorf("scanning into portfolio_group row: %w", err)
 	}
+	if err := checkSizesEquivalent("portfolio group memberhsip", len(mid), len(mca)); err != nil {
```
bcspragu, line 141:
```suggestion
if err := checkSizesEquivalent("portfolio group membership", len(mid), len(mca)); err != nil {
```

frontend/components/portfolio/initiative/membership/MenuButton.vue
```diff
@@ -0,0 +1,162 @@
...
+          </PVButton>
+        </div>
+      </div>
+    </PVOverlayPanel>
+  </div>
+</template>
+
+<style scoped lang="scss">
+// Note, these styles are meant to match the size of the PV checkboxes, which we cannot
+// use directly because of the mixed (dash) state here. They appear to be 1.25rem wide.
```
bcspragu, line 157: Could optionally use variables for this.

frontend/pages/my-data.vue
```diff
@@ -0,0 +1,204 @@
...
+const { fromQueryReactive } = useURLParams()
+const { t } = useI18n()
+const pactaClient = usePACTA()
+const { loading: { withLoading } } = useModal()
+
+const tt = (s: string) => t(`${prefix}.${s}`)
+
+const selectedPortfolioIdsQP = fromQueryReactive('pids')
+const selectedPortfolioGroupIdsQP = fromQueryReactive('pgids')
+const tabQP = fromQueryReactive('tab')
```
bcspragu, line 13: You could update `fromQueryReactive` to take a second, optional parameter that sets a fallback default, would clean up the `(<thing>.value ?? <default>)`s below.

bcspragu (approved):
LGTM!

db/sqldb/migrations/0006_initiative_primary_key.up.sql
bcspragu, line 1: Good catch! Do you expect running this migration to cause any problems (for my own edification when I try to deploy this)?

db/sqldb/portfolio.go
gbdubs: Great question. This wasn't clear to me and took a while to debug. Here's my best explanation for posterity:

- Joins happen before any aggregation or group by.
- 3-table joins will create rows based on the cartesian product of their rows. If we have a primary table (P) and two secondary tables (A) and (B), for a given primary key, if we have 3 values in A (A1, A2, A3), and 2 values in B (B1, B2), then the Cartesian product will have 6 values in the join.
- When we then group by the primary key, that row has six values for both A and B in the join result (A1, A3, A3, A1, A2, A3), (B1, B2, B1, B2, B1, B2).
- We could unwind this through uniqueness sets or similar. However, there are two complicating factors that make this much harder to do: (a) nulls (b) objects that are unbunded into different columns (c) (most difficult) objects that are unbundled into different columns across tables.

After playing around with some row-based solutions, I found this was the simplest way (and probably the most extensible/easy to change: do your group-bys where it's not over a join result, and then just select). This also has the advantage of not generating huge cross products when the cardinality of the subtables is large (as it might be in this case).

db/sqldb/portfolio.go
gbdubs: Excellent catch - this was a leftover of one of my "row munging" strategies from the previous comment which did not work. Removed.

db/sqldb/portfolio_group.go
gbdubs: GC

frontend/components/portfolio/initiative/membership/MenuButton.vue
gbdubs: I'll pass on this for now. However, I'll create a TODO to bring in the RMI PVDesigner codebase in through some kind of linkage/customization scheme, which would make making changes like this easier to handle/upgrade piecemeal. #97

frontend/pages/my-data.vue
gbdubs: Great idea. Done.

gbdubs: This is a huge win - it also allows us to do the implicit clever thing of not setting QPs when they're the defaults, simplifying URLs while not making them any less precise. Thanks for this suggestion!

db/sqldb/migrations/0006_initiative_primary_key.up.sql
gbdubs: Great question. My understanding is that it's essentially equivalent to running a `not null` and `index creation` and `unique` column modification. The not-null constraint should hold based on our bizlogic. The uniquness constraint is the actual source of the need here, so it's plausible that there are duplicate rows. Should that be the case, we could delete those rows when deleting. However, since nobody has used these features in dev yet, we probably won't experience this at all, and if we needed to we could drop the contents of the table to make these hold.

db/sqldb/portfolio.go
bcspragu: >     * 3-table joins will create rows based on the cartesian product of their rows. If we have a primary table (P) and two secondary tables (A) and (B), for a given primary key, if we have 3 values in A (A1, A2, A3), and 2 values in B (B1, B2), then the Cartesian product will have 6 values in the join.

Huh yeah, sure enough:

```sql
CREATE TABLE p (
  id TEXT PRIMARY KEY NOT NULL
);

CREATE TABLE a (
  id TEXT PRIMARY KEY NOT NULL,
  p_id TEXT REFERENCES p (id) NOT NULL
);

CREATE TABLE b (
  id TEXT PRIMARY KEY NOT NULL,
  p_id TEXT REFERENCES p (id) NOT NULL
);


INSERT INTO p (id) VALUES ('p1');
INSERT INTO a (id, p_id) VALUES ('a1', 'p1'), ('a2', 'p1');
INSERT INTO b (id, p_id) VALUES ('b1', 'p1'), ('b2', 'p1'), ('b3', 'p1');

SELECT p.id, ARRAY_AGG(a.id), ARRAY_AGG(b.id)
FROM p
LEFT JOIN a ON p.id = a.p_id
LEFT JOIN b ON p.id = b.p_id
GROUP BY p.id;
```

produces

```
 id |      array_agg      |      array_agg
----+---------------------+---------------------
 p1 | {a2,a1,a2,a1,a2,a1} | {b1,b1,b2,b2,b3,b3}
```

When asking _the internet_ about this, I got the suggestion:

```sql
  SELECT p.id,
      (SELECT ARRAY_AGG(a.id) FROM a WHERE a.p_id = p.id),
      (SELECT ARRAY_AGG(b.id) FROM b WHERE b.p_id = p.id)
  FROM p;
```

Which involves subqueries but otherwise seems like the simplest approach, and I think applies here?

gbdubs: Totally fair, that approach also would work. I think the downside of it is when we have multiple columns from `a` or `b` (and because memberships will often have an `id` and a `created at`, I think this is most cases), a nested query in the select statement then needs to be either (a) assumed order equivalent across multiple expressions, which might be the case or might not, or (b) needs to be multi-selected and the unnested (i.e. put into a composite object, then decomposed).

PR #92, opened 2023-12-27
Backend Infra to Start Analyses

gbdubs:
Creates the core endpoints for starting a run, looking up analysis runs and analysis artifacts. The centerpiece of this PR is the `RunAnalysis` endpoint.

Along the way, the PR makes a few misc cleanups:

- Simplifies the interface of `deleteBlobs`
- Fixes how we were doing enums in oapi so that the generated code generates shared definitions, simplifying and strengthening conversion.
- Looks up the IDs of the artifacts associated with an analysis in the DB, using the query pattern of #91 to avoid incorrect group-by patterns.


cmd/server/pactasrv/analysis.go
```diff
@@ -0,0 +1,343 @@
...
+	if request.Body.Name != nil {
+		mutations = append(mutations, db.SetAnalysisName(*request.Body.Name))
+	}
+	if request.Body.Description != nil {
+		mutations = append(mutations, db.SetAnalysisDescription(*request.Body.Description))
+	}
+	err = s.DB.UpdateAnalysis(s.DB.NoTxn(ctx), id, mutations...)
+	if err != nil {
+		return nil, oapierr.Internal("failed to update analysis", zap.Error(err))
+	}
```
bcspragu, line 89: optional: To avoid loading the analysis twice, you could have `checkAnalysisAuthorization` optionally provide a transaction, and then use a `PutAnalysis`-type DB update instead.

cmd/server/pactasrv/analysis.go
```diff
@@ -0,0 +1,343 @@
...
+	return api.UpdateAnalysis204Response{}, nil
+}
+
+func (s *Server) checkAnalysisAuthorization(ctx context.Context, id pacta.AnalysisID) (*pacta.Analysis, error) {
```
bcspragu, line 93: Strong +1 to the more explicit authorization approach, it's easy to understand + modify, with little room for sneaky (performance + security) footguns

cmd/server/pactasrv/analysis.go
```diff
@@ -0,0 +1,343 @@
...
+	if request.Body.PortfolioId != nil {
+		pID = pacta.PortfolioID(*request.Body.PortfolioId)
+		found++
+	}
+	if found == 0 {
+		return nil, oapierr.BadRequest("one of initiative_id, portfolio_group_id, or portfolio_id is required")
+	}
+	if found > 1 {
+		return nil, oapierr.BadRequest("only one of initiative_id, portfolio_group_id, or portfolio_id may be set")
+	}
```
bcspragu, line 230: Note that these error messages are for us, they don't go to end-users. The end user just gets 400 `Bad Request`

db/sqldb/analysis.go
```diff
@@ -9,25 +9,41 @@ import (
...
+			LEFT JOIN (
+				SELECT 
+					analysis_id,
+					ARRAY_AGG(analysis_artifact.id) AS analysis_artifact_ids
+				FROM analysis_artifact 
+				WHERE analysis_id IN (SELECT id FROM selected_analysis_ids)
+				GROUP BY analysis_id
+			) aas ON aas.analysis_id = analysis.id
+		%[1]s;
+`, where)
```
bcspragu, line 42: Same here, this seems like lots of nested subqueries that we don't need. What's wrong with something like:

```suggestion
	return fmt.Sprintf(`
		SELECT
			analysis.id,
			analysis.analysis_type,
			analysis.owner_id,
			analysis.pacta_version_id,
			analysis.portfolio_snapshot_id,
			analysis.name,
			analysis.description,
			analysis.created_at,
			analysis.ran_at,
			analysis.completed_at,
			analysis.failure_code,
			analysis.failure_message,
			ARRAY_AGG(analysis_artifact.id)
		FROM analysis
		LEFT JOIN analysis_artifact
			ON analysis_artifact.analysis_id = analysis.id
		%[1]s ;; This is the WHERE clause I think. Also, I'd put the `WHERE ` here, so that all the callers don't have to remember it
		GROUP by analysis.id;
	`, where)
```

openapi/pacta.yaml
```diff
@@ -816,6 +952,15 @@ components:
     Language:
       type: string
       enum: *LANGUAGES
+    FileType:
+      type: string
+      enum: *FILE_TYPES
```
bcspragu, line 957: YAML anchors work here, though I would expect using `$ref` to be the more idiomatic approach here, what problem were you seeing with it?

bcspragu (approved):

cmd/server/pactasrv/analysis.go
gbdubs: Generally agree - I think we can handle other cases as they arise. The more I work with this project, the more I like the simplicity of its data model in making Authz trivial. That may change as we get into more advanced listing mechanisms, however.

cmd/server/pactasrv/analysis.go
gbdubs: Great point. I'll file a bug to go through our BadRequest creations and add human-readable error message codes.

gbdubs: #98

db/sqldb/analysis.go
gbdubs: This is mostly for consistency with the other query structure. The other query looked JUST like you're proposing. But then, when I added a second nested relationship table to the query, it broke. The advantage I see of structuring it this way is to prime it so that additional joins will follow the same structure, and be less likely to fall into the pitfall I spent a few hours having to debug.

openapi/pacta.yaml
gbdubs: Sorry I'm actually not sure what you mean here?

cmd/server/pactasrv/analysis.go
gbdubs: I'll skip this at the moment. If we have performance implications that are noticeable I'll obviously reconsider. My thinking is twofold: (a) keeping the authz code as simple as possible has complexity benefits (b) we only allow some fields to be changeable. I like that we expose that set of fields via `db.UpdateFns`, since it allows us to have an explicit API that shows the bizlogic which fields are writable. Put semantics blur this line, and can easily cause a situation where the bizlogic tries to set an field that isn't settable, and ends up surprised by it.

gbdubs:
@bcspragu I'll wait on your response to the two questions above before merging - I want your opinion/perspective!

openapi/pacta.yaml
bcspragu: The thing you're doing with `&THING` and `*THING` is called [YAML 'anchors' and 'aliases'](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/), which is a thing YAML does to allow one to say "take this chunk and pretend it's also here".

What I'm saying is that OpenAPI specs have a mechanism for that, separate from YAML, which is the `$ref` syntax It says broadly the same thing of "take this definition and use it here too". It looks like that's what we were using, so I'm wondering what's wrong with using '$ref' here and just having a single definition for each enum under `/components/`

gbdubs: Beautiful, thank you for the explanation! Makes a ton of sense.

db/sqldb/analysis.go
gbdubs: Copying this here for the discussion: https://github.com/RMI-PACTA/app/pull/91#discussion_r1437297560

openapi/pacta.yaml
gbdubs: Sent #100 to clean this up along with a bunch of other small stuff.

PR #96, opened 2023-12-28
Creates Blob Retrieval Endpoint

gbdubs:
The basic idea here is pretty simple: RMI has indicated to us that metadata is fairly nonsensitive, while the contents of the data (portfolios and reports) is hugely sensitive. My approach to this is to centralize the creation of download URLs in a single place (kind of the opposite of what we did for ADSCI's blob population logic). Additionally, we don't want to populate these proactively because RMI wants audit logs when people actually access the objects. Thus, here's my approach (which this PR starts to implement):

- Have a common download button vue Component, which takes in one or more BlobIDs as the inputs.
- When the download button is clicked, the component starts spinning (loading) it makes a request to our server for a download URL. The button is disabled, spinning during this request. It doesn't block page navigation.
- On the BE, we record (transactionally) audit logs, then generate the download URLs. The ordering ensures our invariant of only granting access when audited. This is the endpoint that this PR implements.
- Then, the download button will transition from spinning into a dial, using `fetch` and `ReadableStream` to increase the percentage done as the download (or downloads) complete.
- Finally the button will turn into a check mark and remain disabled to prevent accidental duplicate downloads.

Thoughts?


cmd/server/pactasrv/blobs.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, err
+	}
+	actorUserID, err := getUserID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	actorIsAdmin, actorIsSuperAdmin, err := s.isAdminOrSuperAdmin(ctx)
+	if err != nil {
+		return nil, err
+	}
```
bcspragu, line 27: Depending on how popular needing this set of info is, I'd add a "gimme the deets" helper that returns all this stuff in a nice `UserInfo` struct or similar.

cmd/server/pactasrv/blobs.go
```diff
@@ -0,0 +1,112 @@
...
+	asMap := map[pacta.BlobID]*pacta.BlobOwnerInformation{}
+	for _, boi := range bois {
+		asMap[boi.BlobID] = boi
+	}
+	auditLogs := []*pacta.AuditLog{}
+	for _, blobID := range blobIDs {
+		boi := asMap[blobID]
+		accessAsOwner := boi.PrimaryTargetOwnerID == actorOwnerID
+		accessAsAdmin := boi.AdminDebugEnabled && actorIsAdmin
+		accessAsSuperAdmin := boi.AdminDebugEnabled && actorIsSuperAdmin
```
bcspragu, line 50: Just a sanity check: even _super_ admins can't look at blobs if the owner didn't opt in? What makes them so super then?

cmd/server/pactasrv/blobs.go
```diff
@@ -0,0 +1,112 @@
...
+			al.ActorType = pacta.AuditLogActorType_User
+		} else if accessAsAdmin {
+			al.ActorType = pacta.AuditLogActorType_Admin
+		} else if accessAsSuperAdmin {
+			al.ActorType = pacta.AuditLogActorType_SuperAdmin
+		} else {
+			// DENY CASE
+			return nil, err404
+		}
+		auditLogs = append(auditLogs, al)
```
bcspragu, line 69: optional: I find code like this easier to understand and modify when the object is constructed once and not changed, e.g.:

```go
var actorType pacta.AuditLogActorType
if accessAsOwner {
	actorType = pacta.AuditLogActorType_User
} else if accessAsAdmin {
	actorType = pacta.AuditLogActorType_Admin
} else if accessAsSuperAdmin {
	actorType = pacta.AuditLogActorType_SuperAdmin
} else {
	// DENY CASE
	return nil, err404
}
auditLogs = append(auditLogs, &pacta.AuditLog{
	Action:             pacta.AuditLogAction_Download,
	ActorType:          actorType,
	ActorID:            string(actorUserID),
	ActorOwner:         &pacta.Owner{ID: actorOwnerID},
	PrimaryTargetType:  boi.PrimaryTargetType,
	PrimaryTargetID:    boi.PrimaryTargetID,
	PrimaryTargetOwner: &pacta.Owner{ID: boi.PrimaryTargetOwnerID},
})
```

I only make this (otherwise pretty pedantic) note because it sounds like this is a pretty sensitive code path.

cmd/server/pactasrv/blobs.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		if db.IsNotFound(err) {
+			return nil, err404
+		}
+		return nil, oapierr.Internal("error getting blobs", zap.Error(err), zap.Strings("blob_ids", asStrs(blobIDs)))
+	}
+
+	err = s.DB.Transactional(ctx, func(tx db.Tx) error {
+		for i, al := range auditLogs {
+			_, err := s.DB.CreateAuditLog(tx, al)
```
bcspragu, line 82: How many blobs are we expecting to load at once on average? Seems like audit log creation would benefit from a batch endpoint, then you could (optionally) drop the transaction.

Edit: looks like one blob at a time on average, so probably not an issue

cmd/server/pactasrv/blobs.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, blob := range blobs {
+		url, err := s.Blob.SignedDownloadURL(ctx, string(blob.BlobURI))
+		if err != nil {
+			return nil, oapierr.Internal("error getting signed download url", zap.Error(err), zap.String("blob_uri", string(blob.BlobURI)))
+		}
+		response.Items = append(response.Items, api.AccessBlobContentRespItem{
+			BlobId:      string(blob.ID),
+			DownloadUrl: url,
+			// TODO(#93) Share source of truth between here and the Azure Blob Storage Library
+			ExpirationTime: s.Now().Add(15 * time.Minute),
```
bcspragu, line 108: I'd update the interface for `SignedDownloadURL` to return the expiration time, otherwise there's some drift here.

pacta/pacta.go
```diff
@@ -260,6 +260,27 @@ func (o *Blob) Clone() *Blob {
 	}
 }
 
+type BlobOwnerInformation struct {
```
bcspragu, line 263: niiiit: `Information` is a long word that ironically adds little information, just `BlobOwner` is probably fine. Even if it means your `bois` variables become a less-entertaining `bos`

bcspragu (approved):
LGTM!

cmd/server/pactasrv/blobs.go
gbdubs: Yep typical case is expected to be one (PR comment describes the "why multiple in the API then" question)

cmd/server/pactasrv/blobs.go
gbdubs: Oh excellent idea, much cleaner than what I was thinking.

cmd/server/pactasrv/blobs.go
gbdubs: SGTM

cmd/server/pactasrv/blobs.go
gbdubs: Yep, this is WAI. Super Admins have other powers (like minting admins, raw access to metadata, etc.)

cmd/server/pactasrv/blobs.go
gbdubs: Great suggestion - done.

PR #99, opened 2023-12-28
Audit Log API Layer

gbdubs:
- Creates API layer for Audit Log Queries (largely mirroring the database layer - I'm thinking of how lovely and useful that approach was for ABOUND, and seeking to replicate it here, albeit for likely a less user-visible use case).
- Tweaks a few things I got wrong in the initial audit log design:
  - we want the `ActorType` to refer to the **pathway through which the user is accessing**, rather than **the property of the user**. This unfortunately required a DB migration to add the new enum values.
  - Adds a `ParseSortBy` method to the DB Layer.
  - Tweaks naming to be more consistent with the core types.

### DIFF
--- a/cmd/server/pactasrv/conv/oapi_to_pacta.go
+++ b/cmd/server/pactasrv/conv/oapi_to_pacta.go
@@ -1,8 +1,10 @@
 package conv
 
 import (
+	"fmt"
 	"regexp"
 
+	"github.com/RMI/pacta/db"
 	"github.com/RMI/pacta/oapierr"
 	api "github.com/RMI/pacta/openapi/pacta"
 	"github.com/RMI/pacta/pacta"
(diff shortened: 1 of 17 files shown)


cmd/server/pactasrv/conv/oapi_to_pacta.go
```diff
@@ -92,3 +94,109 @@ func PortfolioGroupCreateFromOAPI(pg *api.PortfolioGroupCreate, ownerID pacta.Ow
...
+			return nil, fmt.Errorf("converting audit log query where in target type: %w", err)
+		}
+		result.InTargetType = tt
+	}
+	if i.InTargetId != nil {
+		result.InTargetID = *i.InTargetId
+	}
+	if i.InTargetOwnerId != nil {
+		result.InTargetOwnerID = fromStrs[pacta.OwnerID](*i.InTargetOwnerId)
+	}
```
bcspragu, line 153: Did RMI ask for any/all of this functionality? Seems like a lot of ways to slice and dice stuff out of the gate, versus just the MVP.

cmd/server/pactasrv/conv/pacta_to_oapi.go
```diff
@@ -236,3 +236,73 @@ func PortfolioGroupToOAPI(pg *pacta.PortfolioGroup) (*api.PortfolioGroup, error)
 func PortfolioGroupsToOAPI(pgs []*pacta.PortfolioGroup) ([]*api.PortfolioGroup, error) {
 	return convAll(pgs, PortfolioGroupToOAPI)
 }
+
+func auditLogActorTypeToOAPI(i pacta.AuditLogActorType) (api.AuditLogActorType, error) {
+	return api.AuditLogActorType(string(i)), nil
```
bcspragu, line 241: Does the generated `api` code not produce unique enums? Seems like it'd be better to explicitly map them, instead of implicitly expecting them to stay in sync. Same below

bcspragu (approved):

cmd/server/pactasrv/conv/oapi_to_pacta.go
gbdubs: I had a long convo with Hodie about this early in the project. My recollection is that the goal was to make audit logs useful for debuging + analytics, in addition to serving their purpose of giving folks confidence in the data handling practices.

gbdubs: Another benefit is that the API exposed maps cleanly onto the PV Pagination Table interface (sorting, filtering etc) so building a maintainable frontend for this is going to be delightfully easy.

cmd/server/pactasrv/conv/pacta_to_oapi.go
gbdubs: SGTM, done.

gbdubs: Great call, found several small typos this way.

PR #100, opened 2024-01-02
Various Enum Fixes + Tweaks

gbdubs:
- Adds an enum prefix for every OpenAPI enum type, to avoid collisions + clean up the namespace (otherwise constants like `api.En` are generated).
- Removes use of YAML feature I hadn't understood prior (anchors + links), putting everything into the `schemas`, since `definitions` look like they're a Swagger (OAPI2.0) construct.
- Adds some basic Enum Round-Trip unit tests to pactasrv/conv
- removes usage of `pacta.ParseX` functions within the conv package, since these are less reliable than using the one for one, switch driven converters.

### DIFF
--- a/cmd/server/pactasrv/conv/oapi_to_pacta.go
+++ b/cmd/server/pactasrv/conv/oapi_to_pacta.go
@@ -124,25 +124,25 @@ func PortfolioGroupCreateFromOAPI(pg *api.PortfolioGroupCreate, ownerID pacta.Ow
 
 func auditLogActionFromOAPI(i api.AuditLogAction) (pacta.AuditLogAction, error) {
 	switch i {
-	case api.AuditLogActionCREATE:
+	case api.AuditLogActionCreate:
 		return pacta.AuditLogAction_Create, nil
-	case api.AuditLogActionUPDATE:
+	case api.AuditLogActionUpdate:
 		return pacta.AuditLogAction_Update, nil
-	case api.AuditLogActionDELETE:
+	case api.AuditLogActionDelete:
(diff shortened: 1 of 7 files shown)


cmd/server/pactasrv/conv/oapi_to_pacta.go
```diff
@@ -109,15 +123,65 @@ func PortfolioGroupCreateFromOAPI(pg *api.PortfolioGroupCreate, ownerID pacta.Ow
...
+	switch i {
+	case api.AuditLogActionCREATE:
+		return pacta.AuditLogAction_Create, nil
+	case api.AuditLogActionUPDATE:
+		return pacta.AuditLogAction_Update, nil
+	case api.AuditLogActionDELETE:
+		return pacta.AuditLogAction_Delete, nil
+	case api.AuditLogActionADDTO:
+		return pacta.AuditLogAction_AddTo, nil
+	case api.AuditLogActionREMOVEFROM:
```
bcspragu, line 135: nit: CANYOUADDSOMEUNDERSCORESHERE? OTHERWISEITSHARDTOREAD

ANDBELOW

(let me know if its annoying to do, I should be able to do it in one ~5s pass in a text editor)

bcspragu (approved):

cmd/server/pactasrv/conv/oapi_to_pacta.go
gbdubs: Went with CamelCase - couldn't figure out how to configure (globally) enum code generation.

PR #102, opened 2024-01-02
Re-run `generate:openapi` to bring frontend up to speed

gbdubs:
In the last week or so I've done a ton of BE changes, which haven't yet made it to the frontend. This PR regenerates the openapi typings and tweaks the features that have been impacted by the API changes (exclusively language + enum labeling). 

This is separated out as a simple PR to avoid huge diffs in a bizlogic-changing PR down the road.


bcspragu (approved):
LGTM, thanks for putting this in it's own PR, makes it super easy to review

PR #117, opened 2024-01-05
Creates Backend for Merging User Accounts

gbdubs:
User Merge is apparently a frequent support flow, and required some delicate thinking around audit logs. This PR implements the backend for it.

Additional changes:
- Adds a new audit log type for `TRANSFER_OWNERSHIP`
- Creates a `DeleteOwner` DB method
- Updates `DeleteUser` to do the recursive deletion of their owned assets + return blobURIs
- Updates `/delete-user` to delete the downstream blobs.

### DIFF
--- a/cmd/server/pactasrv/admin.go
+++ b/cmd/server/pactasrv/admin.go
@@ -1,0 +1,159 @@
+package pactasrv
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/RMI/pacta/db"
+	"github.com/RMI/pacta/oapierr"
+	api "github.com/RMI/pacta/openapi/pacta"
+	"github.com/RMI/pacta/pacta"
+	"go.uber.org/zap"
+)
+
+// Merges two users together
+// (POST /admin/merge-users)
+func (s *Server) MergeUsers(ctx context.Context, request api.MergeUsersRequestObject) (api.MergeUsersResponseObject, error) {
+	req := request.Body
+	actorUserInfo, err := s.getActorInfoOrFail(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !actorUserInfo.IsAdmin && !actorUserInfo.IsSuperAdmin {
+		return nil, oapierr.Forbidden("only admins can merge users",
+			zap.String("actor_user_id", string(actorUserInfo.UserID)),
+			zap.String("from_user_id", req.FromUserId),
+			zap.String("to_user_id", req.ToUserId))
+	}
+
+	sourceUID := pacta.UserID(req.FromUserId)
+	destUID := pacta.UserID(req.ToUserId)
+
+	var (
+		numIncompleteUploads, numAnalyses, numPortfolios, numPortfolioGroups, numAuditLogs, numAuditLogsCreated int
+		buris                                                                                                   []pacta.BlobURI
+	)
+
+	err = s.DB.Transactional(ctx, func(tx db.Tx) error {
+		sourceOwner, err := s.DB.GetOwnerForUser(tx, sourceUID)
+		if err != nil {
+			return fmt.Errorf("failed to get owner for source user: %w", err)
+		}
+		destOwner, err := s.DB.GetOwnerForUser(tx, destUID)
+		if err != nil {
+			return fmt.Errorf("failed to get owner for destination user: %w", err)
+		}
+
+		// Note we do an audit log transfer FIRST so that we don't transfer the audit logs generated from the transfer itself.
+		nal, err := s.DB.TransferAuditLogOwnership(tx, sourceUID, destUID, sourceOwner, destOwner)
+		if err != nil {
+			return fmt.Errorf("failed to transfer audit log ownership: %w", err)
+		}
+		numAuditLogs = nal
+
+		auditLogsToCreate := []pacta.AuditLog{}
+		addAuditLog := func(t pacta.AuditLogTargetType, id string) {
+			auditLogsToCreate = append(auditLogsToCreate, pacta.AuditLog{
+				Action:               pacta.AuditLogAction_TransferOwnership,
+				ActorType:            pacta.AuditLogActorType_Admin,
+				ActorID:              string(actorUserInfo.UserID),
+				ActorOwner:           &pacta.Owner{ID: actorUserInfo.OwnerID},
+				PrimaryTargetType:    t,
+				PrimaryTargetID:      id,
+				PrimaryTargetOwner:   &pacta.Owner{ID: destOwner},
+				SecondaryTargetType:  pacta.AuditLogTargetType_User,
+				SecondaryTargetID:    string(sourceUID),
+				SecondaryTargetOwner: &pacta.Owner{ID: sourceOwner},
+			})
+		}
+
+		incompleteUploads, err := s.DB.IncompleteUploadsByOwner(tx, sourceOwner)
+		if err != nil {
+			return fmt.Errorf("failed to get incomplete uploads for source owner: %w", err)
+		}
+		for i, upload := range incompleteUploads {
+			err := s.DB.UpdateIncompleteUpload(tx, upload.ID, db.SetIncompleteUploadOwner(destOwner))
+			if err != nil {
+				return fmt.Errorf("failed to update upload owner %d/%d: %w", i, len(incompleteUploads), err)
+			}
+			addAuditLog(pacta.AuditLogTargetType_IncompleteUpload, string(upload.ID))
+		}
+		numIncompleteUploads = len(incompleteUploads)
+
+		analyses, err := s.DB.AnalysesByOwner(tx, sourceOwner)
+		if err != nil {
+			return fmt.Errorf("failed to get analyses for source owner: %w", err)
+		}
+		for i, analysis := range analyses {
+			err := s.DB.UpdateAnalysis(tx, analysis.ID, db.SetAnalysisOwner(destOwner))
+			if err != nil {
+				return fmt.Errorf("failed to update analysis owner %d/%d: %w", i, len(analyses), err)
+			}
+			addAuditLog(pacta.AuditLogTargetType_Analysis, string(analysis.ID))
+		}
+		numAnalyses = len(analyses)
+
+		portfolios, err := s.DB.PortfoliosByOwner(tx, sourceOwner)
+		if err != nil {
+			return fmt.Errorf("failed to get portfolios for source owner: %w", err)
+		}
+		for i, portfolio := range portfolios {
+			err := s.DB.UpdatePortfolio(tx, portfolio.ID, db.SetPortfolioOwner(destOwner))
+			if err != nil {
+				return fmt.Errorf("failed to update portfolio owner %d/%d: %w", i, len(portfolios), err)
+			}
+			addAuditLog(pacta.AuditLogTargetType_Portfolio, string(portfolio.ID))
+		}
+		numPortfolios = len(portfolios)
+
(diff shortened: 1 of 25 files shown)


cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+		numPortfolioGroups = len(portfolioGroups)
+
+		for _, auditLog := range auditLogsToCreate {
+			_, err := s.DB.CreateAuditLog(tx, &auditLog)
+			if err != nil {
+				return fmt.Errorf("failed to create audit log: %w", err)
+			}
+		}
+		numAuditLogsCreated = len(auditLogsToCreate)
+
```
gbdubs, line 129: Option I'm considering: we could do a lookup (by owner + by user) of each of the entity groups just transferred to ensure they're no longer present (similar to what I did at the end of the TXN to transfer audit log ownership). I'd like to hear your thoughts on whether that's a good idea - particularly because the db.DeleteUser call will then delete these elements if they don't transfer.

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+	api "github.com/RMI/pacta/openapi/pacta"
+	"github.com/RMI/pacta/pacta"
+	"go.uber.org/zap"
+)
+
+// Merges two users together
+// (POST /admin/merge-users)
+func (s *Server) MergeUsers(ctx context.Context, request api.MergeUsersRequestObject) (api.MergeUsersResponseObject, error) {
+	req := request.Body
+	actorUserInfo, err := s.getActorInfoOrFail(ctx)
```
bcspragu, line 18: nit: The `OrFail` seems implicit in the idea that it has a `(_, error)` return signature, no? How about just `s.getActorInfo(ctx)`

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if !actorUserInfo.IsAdmin && !actorUserInfo.IsSuperAdmin {
+		return nil, oapierr.Forbidden("only admins can merge users",
+			zap.String("actor_user_id", string(actorUserInfo.UserID)),
+			zap.String("from_user_id", req.FromUserId),
+			zap.String("to_user_id", req.ToUserId))
+	}
+
+	sourceUID := pacta.UserID(req.FromUserId)
+	destUID := pacta.UserID(req.ToUserId)
```
bcspragu, line 30: I'm going to make a dumb suggestion: check that these don't match. If they do, I'd either throw an error (preferred) or return OK (i.e a no-op). Since this is an admin/support endpoint, I expect weird input from it being manually run, and I think `src == dest` will otherwise do weird things.

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get owner for destination user: %w", err)
+		}
+
+		// Note we do an audit log transfer FIRST so that we don't transfer the audit logs generated from the transfer itself.
+		nal, err := s.DB.TransferAuditLogOwnership(tx, sourceUID, destUID, sourceOwner, destOwner)
+		if err != nil {
+			return fmt.Errorf("failed to transfer audit log ownership: %w", err)
+		}
+		numAuditLogs = nal
```
bcspragu, line 52: I'd consider not moving existing audit logs, it seems tricky and error-prone, and breaks the idea that audit logs are "write-only".

If it's important to show the merged user info in the UI for this stuff, I'd record merges in a separate table (also without FKCs), and then just do that mapping when users look at the audit logs. A bit more work on the read-side, but it's a straightforward mapping and it removes a bunch of nuance around audit logs.

Instead of a separate table, you _could_ use the `TRANSFER_OWNERSHIP` logs to get the same info, though that make audit logs "load-bearing"

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		incompleteUploads, err := s.DB.IncompleteUploadsByOwner(tx, sourceOwner)
+		if err != nil {
+			return fmt.Errorf("failed to get incomplete uploads for source owner: %w", err)
+		}
+		for i, upload := range incompleteUploads {
+			err := s.DB.UpdateIncompleteUpload(tx, upload.ID, db.SetIncompleteUploadOwner(destOwner))
+			if err != nil {
+				return fmt.Errorf("failed to update upload owner %d/%d: %w", i, len(incompleteUploads), err)
```
bcspragu, line 77: nit: This log (and below) will be confusing because `i` is zero-indexed and `len(...)` is 1-indexed, e.g you'll always get `failed to update ... N-1/N` for the last one.

But also since there's no order on the load call, the indexes aren't particularly useful for debugging, so I'd use IDs instead.

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+			err := s.DB.UpdatePortfolioGroup(tx, portfolioGroup.ID, db.SetPortfolioGroupOwner(destOwner))
+			if err != nil {
+				return fmt.Errorf("failed to update portfolio group owner %d/%d: %w", i, len(portfolioGroups), err)
+			}
+			addAuditLog(pacta.AuditLogTargetType_PortfolioGroup, string(portfolioGroup.ID))
+		}
+		numPortfolioGroups = len(portfolioGroups)
+
+		for _, auditLog := range auditLogsToCreate {
+			_, err := s.DB.CreateAuditLog(tx, &auditLog)
```
bcspragu, line 123: optional: seems like `auditLogsToCreate` might actually have a fairly large cardinality, consider making a `BatchCreateAuditLog` DB endpoint

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to create audit log: %w", err)
+			}
+		}
+		numAuditLogsCreated = len(auditLogsToCreate)
+
+		// Now that we've transferred all the audit logs, we can delete the user.
```
bcspragu, line 130: Is this supposed to say "transferred all the entities"? Audit logs don't actually use FKCs (which is good/correct), so they should have no bearing on deleting the user.

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+		newBuris, err := s.DB.DeleteUser(tx, sourceUID)
+		if err != nil {
+			return fmt.Errorf("failed to delete user: %w", err)
+		}
+		buris = append(buris, newBuris...)
```
bcspragu, line 135: It looks like the only blobs this could return come from deleting things the source user would have been the owner of, which should be zero things now. Instead of deleting these blobs, I'd check that there's no blobs, and log a stern error if there are for some reason. Deleting blobs that shouldn't exist seems like it could cause data loss.

cmd/server/pactasrv/admin.go
```diff
@@ -0,0 +1,159 @@
...
+		zap.String("from_user_id", req.FromUserId),
+		zap.String("to_user_id", req.ToUserId),
+		zap.Int("num_audit_logs_transferred", numAuditLogs),
+		zap.Int("num_incomplete_uploads", numIncompleteUploads),
+		zap.Int("num_analyses", numAnalyses),
+		zap.Int("num_portfolios", numPortfolios),
+		zap.Int("num_portfolio_groups", numPortfolioGroups),
+		zap.Int("num_audit_logs_created", numAuditLogsCreated),
+	)
+	return api.MergeUsers204Response{}, nil
```
bcspragu, line 158: optional: Could return the metrics we're logging here in the response, as a sanity check for the admin running it.

db/sqldb/audit_log.go
```diff
@@ -88,6 +88,70 @@ func (d *DB) CreateAuditLog(tx db.Tx, a *pacta.AuditLog) (pacta.AuditLogID, erro
...
+			SELECT 
+				COUNT(*)
+			FROM audit_log
+			WHERE 
+				actor_id = $1 OR
+				primary_target_id = $1 OR
+				secondary_target_id = $1 OR
+				actor_owner_id = $2 OR
+				primary_target_owner_id = $2 OR
+				secondary_target_owner_id = $2;`, sourceUserID, sourceOwnerID)
```
bcspragu, line 105: I think you want to put a `... FOR UPDATE` at the end here (at least for the first invocation, not sure about the second), e.g. `SELECT ... FROM ... WHERE ... FOR UPDATE`, see [1] [2]

[1] https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS
[2] https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE

bcspragu (approved):
LGTM, main thing is that I would not transfer audit records as part of the merge, or really ever modify them period.

cmd/server/pactasrv/admin.go
gbdubs: Excellent suggestion. Done.

cmd/server/pactasrv/admin.go
gbdubs: Good point this is ambiguous.

The "OrFail" is intended to capture "if anonymous what should we do", because I don't want us to handle a logged out user and a user lookup failure the same. I'll change to "OrErrIfAnon"

cmd/server/pactasrv/admin.go
gbdubs: This is a SUPERB suggestion.

cmd/server/pactasrv/admin.go
gbdubs: The reason I tend to like including indexes for failure conditions within potentially large ops is that it helps distinguish between "this fails all the time" vs "this fails at large scales" vs "this fails intermittently". The index helps discern that, and if we don't have it in the logs, the logs are mute at distinguishing between these three cases.

However, Great Catch on the fencepost err.

cmd/server/pactasrv/admin.go
gbdubs: Great call - done.

cmd/server/pactasrv/admin.go
gbdubs: Good catch - yep just a typo.

cmd/server/pactasrv/admin.go
gbdubs: This is another SUPERB suggestion. Done.

cmd/server/pactasrv/admin.go
gbdubs: I generally agree that editing audit logs is an unstable paradigm. 

I think what you propose is also dangerous however, because it introduces much less stable semantics around how audit logs can be accessed and retrieved, which are also an important security end. My thought with doing the data modification in place is that we only have to navigate the complexity of "who should get access to what" at exactly one juncture, and then afterwards queries and reads and creates all need no modification. 

I agree using TRANSFER_OWNERSHIP logs is a nightmare which we don't want to descend into. 

Not sure how to resolve/move forward on this one.

db/sqldb/audit_log.go
gbdubs: Mooted now that we don't do an update.

gbdubs:
> LGTM, main thing is that I would not transfer audit records as part of the merge, or really ever modify them period.

Discussed this offline and was sufficently convinced that it's important for us to never edit these. Added some new mechanisms for recursively expanding the set of users used for a query, which will allow audit logs to be visible post-merge, through some magic.

PR #122, opened 2024-01-07
Full Authz + Audit Log Implementation

gbdubs:
This PR adds audit logging and authorization to all `oapi` endpoints, using a series of iterated-upon utilities. Some notable things:

- I went through like 5 refactors of this code during writing it, so if you have structural comments, let's talk about them in person. The many roads not traveled are interesting.
- Centralized utilities for simplifying and making authorization code more easy to grok is in `authz`.
- I tried hard to get authorization on a type-by-type basis sorted into a single, linear, readable method, with exactly the same pattern each time: GetActorInfo, LookUpData, StructuredIdentificationInfo, ActionSpecificAuthorizationLogic, StandardizedHandleAuditLogAndReject.
- This approach probably generates more audit logs than we want in a ready-for-release version. However, I figured it's actually easier to build out audit logging on a 1:1 basis with authorization than it is to parse that apart on a case by case basis. I expect we'll tweak where audit logs are generated over time, but this PR provides the infrastructure for them everywhere.
- I've separated this PR (generating audit logs + adding authorization) from a coming PR on displaying audit logs on the front-end. I think the separation is clean, but I apologize if there are stragglers.


cmd/server/pactasrv/analysis.go
```diff
@@ -341,3 +283,88 @@ func (s *Server) RunAnalysis(ctx context.Context, request api.RunAnalysisRequest
 
 	return api.RunAnalysis200JSONResponse{AnalysisId: string(result)}, nil
 }
+
+func (s *Server) analysisDoAuthzAndAuditLog(ctx context.Context, analysisID pacta.AnalysisID, action pacta.AuditLogAction) error {
```
bcspragu, line 287: optional: All these helpers could return the entity they've loaded, saves you a DB call on the GET path.

cmd/server/pactasrv/analysis.go
```diff
@@ -341,3 +283,88 @@ func (s *Server) RunAnalysis(ctx context.Context, request api.RunAnalysisRequest
...
+	as := &authzStatus{
+		primaryTargetID:      string(analysisID),
+		primaryTargetType:    pacta.AuditLogTargetType_Analysis,
+		primaryTargetOwnerID: analysis.Owner.ID,
+		actorInfo:            actorInfo,
+		action:               action,
+	}
+	switch action {
+	case pacta.AuditLogAction_Update, pacta.AuditLogAction_Delete, pacta.AuditLogAction_ReadMetadata:
+		as.isAuthorized, as.authorizedAsActorType = allowIfAdminOrOwner(actorInfo, analysis.Owner.ID)
```
bcspragu, line 308: I think the approach of handling authz and audit logging at the same time is fine, but using the audit logging _verbs_ for authz seems like muddying the concepts too far. I'm not suggesting changing it, but I'd just be aware of it as you go forward, I can imagine areas where this could be problematic.

If you hit those cases, I'd just separate out the authz verb from the audit log verb.

cmd/server/pactasrv/analysis.go
```diff
@@ -341,3 +283,88 @@ func (s *Server) RunAnalysis(ctx context.Context, request api.RunAnalysisRequest
...
+	default:
+		return fmt.Errorf("unknown action %q for analysis authz", action)
+
+	}
+	return s.auditLogIfAuthorizedOrFail(ctx, as)
```
bcspragu, line 313: nit: Same note as in the previous PR, where the `OrFail` seems redundant, doubly so with the `IfAuthorized`.

cmd/server/pactasrv/analysis.go
```diff
@@ -341,3 +283,88 @@ func (s *Server) RunAnalysis(ctx context.Context, request api.RunAnalysisRequest
...
+			as.isAuthorized, as.authorizedAsActorType = false, nil
+		}
+	case pacta.AuditLogAction_EnableAdminDebug,
+		pacta.AuditLogAction_DisableAdminDebug,
+		pacta.AuditLogAction_EnableSharing,
+		pacta.AuditLogAction_DisableSharing:
+		as.isAuthorized, as.authorizedAsActorType = allowIfOwner(actorInfo, analysis.Owner.ID)
+	case pacta.AuditLogAction_Update,
+		pacta.AuditLogAction_Delete:
+		as.isAuthorized, as.authorizedAsActorType = allowIfAdminOrOwner(actorInfo, analysis.Owner.ID)
```
bcspragu, line 365: Looks like admins are able to update/delete analysis artifacts regardless of `AdminDebugEnabled`? Same with a few other resources it looks like, is this intended behavior?

cmd/server/pactasrv/authz.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type authzStatus struct {
+	primaryTargetID      string
+	primaryTargetType    pacta.AuditLogTargetType
+	primaryTargetOwnerID pacta.OwnerID
+
+	secondaryTargetID      *string
+	secondaryTargetType    *pacta.AuditLogTargetType
+	secondaryTargetOwnerID *pacta.OwnerID
```
bcspragu, line 18: optional: Instead of using pointers here, which has awkward semantics, I'd just use empty strings to denote "empty", it's the Go-y way to do these things. We never have a valid ID/TargetType that's a blank string.

cmd/server/pactasrv/authz.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func allowIfOwner(actorInfo actorInfo, targetOwnerID pacta.OwnerID) (bool, *pacta.AuditLogActorType) {
+	if actorInfo.OwnerID == targetOwnerID {
+		return true, ptr(pacta.AuditLogActorType_Owner)
+	}
+	return false, nil
+}
+
+func allowIfAdminOrOwner(actorInfo actorInfo, targetOwnerID pacta.OwnerID) (bool, *pacta.AuditLogActorType) {
```
bcspragu, line 126: optional: to avoid duplicating code, you could compose this of `allowIfOwner` and `allowIfAdmin`

cmd/server/pactasrv/blobs.go
```diff
@@ -22,23 +22,23 @@ func (s *Server) AccessBlobContent(ctx context.Context, request api.AccessBlobCo
...
-	bos, err := s.DB.BlobContexts(s.DB.NoTxn(ctx), blobIDs)
+	bcs, err := s.DB.BlobContexts(s.DB.NoTxn(ctx), blobIDs)
 	if err != nil {
 		if db.IsNotFound(err) {
 			return nil, err404
 		}
 		return nil, oapierr.Internal("error getting blob owners", zap.Error(err), zap.Strings("blob_ids", asStrs(blobIDs)))
 	}
 	asMap := map[pacta.BlobID]*pacta.BlobContext{}
-	for _, boi := range bos {
```
bcspragu, line 33: https://www.youtube.com/watch?v=WZpLQa_cDKE

cmd/server/pactasrv/initiative_invitation.go
```diff
@@ -121,3 +133,65 @@ func (s *Server) ListInitiativeInvitations(ctx context.Context, request api.List
...
+	iurs, err := s.DB.InitiativeUserRelationshipsByInitiative(s.DB.NoTxn(ctx), iID)
+	if err != nil {
+		return oapierr.Internal("failed to list initiative user relationships", zap.Error(err))
+	}
+	actorIsInitiativeManager := false
+	for _, iur := range iurs {
+		if iur.User.ID == actorInfo.UserID && iur.Manager {
+			actorIsInitiativeManager = true
+			break
+		}
```
bcspragu, line 171: optional nit: if there's only one `iur` per user, you can exit early if access is granted or denied

```suggestion
		if iur.User.ID == actorInfo.UserID {
			actorIsInitiativeManager = iur.Manager
			break
		}
```

cmd/server/pactasrv/pactasrv.go
```diff
@@ -211,23 +215,37 @@ type actorInfo struct {
...
 		UserID:       actorUserID,
 		OwnerID:      actorOwnerID,
 		IsAdmin:      actorIsAdmin,
 		IsSuperAdmin: actorIsSuperAdmin,
 	}, nil
 }
+
+func (s *Server) getActorInfoOrAnon(ctx context.Context) (actorInfo, error) {
+	_, err := getUserID(ctx)
+	if err != nil {
```
bcspragu, line 243: _technically_ there are other reasons we can fail to get the user ID from the context, but calling the user 'anonymous' regardless is probably fine

bcspragu (approved):
LGTM, excellent auth system! My main reservation was reusing audit log verbs for authz, but after going through the whole PR, I can see how annoying it would be to have a second set of verbs for authz, seems like a pragmatic shortcut.

cmd/server/pactasrv/analysis.go
gbdubs: I went back on this SO MANY TIMES lol. I tend to agree with you, particularly for big systems, couplings between auth and logging are bound to cause issues. However, the thing I kept coming back to was accidental drift: thinking "hm what audit verb should this be" for an authz verb that didn't fit precicely into an existing one. I eventually came to the conclusion that if we have two sets, we invite drift between the idea of "check authorization" and "log successful authorization", and after a few such accidents, I realized this is probably less error prone for a (relatively) small project like this one.

cmd/server/pactasrv/analysis.go
gbdubs: Yes! Admins can update object metadata, and delete objects, they just cannot **see the actual content of portfolios**, which is why we isolate the checks for ADE to (a) blob => download URL endpoints, and (b) changing of ADE status.

cmd/server/pactasrv/authz.go
gbdubs: SGTM.

cmd/server/pactasrv/analysis.go
gbdubs: I thought about that, and ultimately decided against it just for simplicity - If we end up seeing user-visible performance as a result I'm totally up for it, but I love the simplicity of the same method signature without a few `_, _, _, err := ...` floating around (which would be the case to obviate things like membership lookups)

cmd/server/pactasrv/analysis.go
gbdubs: Fixed (in a hellish merge :))

cmd/server/pactasrv/authz.go
gbdubs: GC. I'd prefer not to introspect the result of `allowIfOwner` before falling through to `allowIfAdmin`, but good call on the second half, so I've changed the fallback to be allowIfAdmin:

```
func allowIfAdminOrOwner(actorInfo actorInfo, targetOwnerID pacta.OwnerID) (bool, *pacta.AuditLogActorType) {
	if actorInfo.OwnerID == targetOwnerID {
		return true, ptr(pacta.AuditLogActorType_Owner)
	}
	return allowIfAdmin(actorInfo)
}
```

cmd/server/pactasrv/blobs.go
gbdubs: LOL I realized I'd neglected to undo my masterpiece

cmd/server/pactasrv/initiative_invitation.go
gbdubs: good call.

cmd/server/pactasrv/pactasrv.go
gbdubs: Yeah it's definitely an edge case - I doubt we'll hit it more than a handful of times over the operation of the site - but if/when user lookup fails, that really should be handled differently than an anonymous user to the site, which is why this differentiation feels important. It's the `(*x, err)` pattern of go after all!

PR #101, opened 2024-01-02
Early Review: Running Analysis

gbdubs:
Looking for early feedback here - does this look like a reasonable request structure, and if so, have I interpreted how to run the report correctly? I want validation of that before I head too far down this road!

PR #129, opened 2024-01-09
Running Analysis

bcspragu:
This is basically #101 rebased onto main, with some small refactorings applied and some deps updated.

You can see my tweaks more or less in isolation in this commit: https://github.com/RMI-PACTA/app/commit/95c31124cf33bbab1c4664dac69e713ca3de5976

I'm going to do some testing (against local Docker + Azure) at this PR, and make any larger tweaks in a follow up PR


gbdubs (approved):
Sweet - glad this proved workable!

PR #130, opened 2024-01-10
Use a single Azure topic for all our async notifications

bcspragu:
Managing the machinery for three topics is too annoying, and it's easier and quicker and generally better to do this than set up all the local + deployed resources to make N topics work.

Instead, we use the `EventType` of a notification to indicate, well, the type of the event (e.g. `parsed-portfolio`, `created-report`, etc)

I'll send out the associated Terraform changes momentarily

### DIFF
--- a/azure/azevents/azevents.go
+++ b/azure/azevents/azevents.go
@@ -33,9 +33,7 @@ type Config struct {
 	// prevent random unauthenticated internet requests from triggering webhooks.
 	AllowedAuthSecrets []string
 
-	ParsedPortfolioTopicName string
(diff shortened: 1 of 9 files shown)


bcspragu:
The main downside is that I've taken out our nice generic machinery, because we don't know the type when we parse the outer event message, so we use `json.RawMessage` to delay parsing, which makes it a runtime thing instead.

gbdubs (approved):
SGTM, I ran into this same bug this morning when working on a different PR. Thanks for the fix!

PR #131, opened 2024-01-10
Allows Computed URL Parameters to not overwrite one another

gbdubs:
Discussed offline the need for something like this, @bcspragu  had the bright idea to use `watch` on the current route, which ended up being the thing that clinched this working.

We'll be able to back-export this to a bunch of other projects once agreed upon!

### DIFF
--- a/frontend/composables/useURLParams.ts
+++ b/frontend/composables/useURLParams.ts
@@ -6,7 +6,49 @@ export const useURLParams = () => {
   const route = useRoute()
   const router = useRouter()
 
+  // A common bug prior to the pending-values approach was that within a single tick, we'd do multiple updates like
+  // (1) setVal('foo', 1)
+  // (2) setVal('bar', 2)
+  // **router.currentRoute** is a reactive object that **does not get updated between ticks**.
+  // Thus setVal('foo', 1) would set the value in the URL, but then setVal('bar', 2) would overwrite it.
+  // Said another way, `setVal('foo', 1)` immediately followed by `getVal('foo'), would return the original value.
+  // The PendingValues approach solves this by tracking the values that are pending to be set in the URL,
+  // **and using them as the source of truth when set for `getVal` queries**.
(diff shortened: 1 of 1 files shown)


frontend/composables/useURLParams.ts
```diff
@@ -49,7 +105,9 @@ export const useURLParams = () => {
   const fromQueryReactiveWithDefault = (key: string, def: string): WritableComputedRef<string> => {
     const fqr = fromQueryReactive(key)
     return computed({
-      get: () => fqr.value ?? def,
+      get: () => {
```
bcspragu, line 108: Looks like this change can probably be undone

frontend/composables/useURLParams.ts
```diff
@@ -25,18 +67,32 @@ export const useURLParams = () => {
...
-  const setVal = (key: string, val: string | undefined) => {
+  const resolvePendingValues = () => {
+    const pvs = pendingValues.value
     const query = new URLSearchParams(stringifyQuery(router.currentRoute.value.query))
-    if (val) {
-      query.set(key, val)
-    } else {
-      query.delete(key)
+    for (const [key, val] of pvs) {
+      if (val) {
```
bcspragu, line 74: I'd explicitly check for `val !== undefined` here, blank strings seem like a sharp edge here.

bcspragu (approved):
LGTM, though I must admit that I find this completely unintelligible, between the nested `watch`s and the general async-ness of things, but it's library code, so it's fine if it's unintelligible if it makes the other code more ergonomic + clear.

I'd test this thoroughly under all the scenarios you can think of before porting this to other projects.

frontend/composables/useURLParams.ts
gbdubs: GC

frontend/composables/useURLParams.ts
gbdubs: Excellent catch.

PR #137, opened 2024-01-11
First Draft of Audit Log FE

gbdubs:
Gets a first draft of the audit log functionality into the frontend to make testing Audit Log testing easier.

Notable outstanding TODOs:

- support for `where` queries
- refresh on sort change
- hiding column selection in a sub menu or something

### DIFF
--- a/frontend/lib/auditlogquery/index.ts
+++ b/frontend/lib/auditlogquery/index.ts
@@ -1,6 +1,6 @@
 import { type AuditLogQueryReq, type AuditLogQuerySort, type AuditLogQueryWhere, type AuditLogQuerySortBy } from '@/openapi/generated/pacta'
 import { type WritableComputedRef } from 'vue'
-import { type useLocalePath } from '@nuxtjs/i18n/dist/runtime/composables'
+import { type LocalePathFunction } from 'vue-i18n-routing'
 import { computed } from 'vue'
 
 const encodeAuditLogQuerySorts = (sorts: AuditLogQuerySort[]): string => {
@@ -176,7 +176,7 @@ export const urlReactiveAuditLogQuery = (fromQueryReactiveWithDefault: (key: str
   })
 }
 
-export const createURLAuditLogQuery = (localePath: ReturnType<typeof useLocalePath>, req: AuditLogQueryReq): string => {
+export const createURLAuditLogQuery = (localePath: LocalePathFunction, req: AuditLogQueryReq): string => {
   const qSorts = encodeAuditLogQuerySorts(req.sorts ?? [])
   const qWheres = encodeAuditLogQueryWheres(req.wheres)
   const qLimit = encodeAuditLogQueryLimit(req.limit ?? limitDefault)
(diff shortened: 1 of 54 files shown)


frontend/lib/auditlogquery/index.ts
```diff
@@ -0,0 +1,200 @@
...
+    set: (value: AuditLogQueryReq) => {
+      qSorts.value = encodeAuditLogQuerySorts(value.sorts ?? [])
+      qWheres.value = encodeAuditLogQueryWheres(value.wheres)
+      qLimit.value = encodeAuditLogQueryLimit(value.limit ?? limitDefault)
+      qCursor.value = encodeAuditLogQueryCursor(value.cursor ?? '')
+    },
+  })
+}
+
+export const createURLAuditLogQuery = (localePath: ReturnType<typeof useLocalePath>, req: AuditLogQueryReq): string => {
```
bcspragu, line 179: Looks like `ReturnType<typeof useLocalePath>` can just be `LocalePathFunction`? At least if I followed that chain right

frontend/lib/auditlogquery/index.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+const sortsQP = 's'
+const wheresQP = 'w'
+const limitQP = 'l'
+const limitDefault = 100
+const cursorQP = 'c'
+const pageURLBase = '/auditlog'
+
+export const urlReactiveAuditLogQuery = (fromQueryReactiveWithDefault: (key: string, defaultValue: string) => WritableComputedRef<string>): WritableComputedRef<AuditLogQueryReq> => {
```
bcspragu, line 155: You're playing a dangerous game, making library functions that depend on access to things that use `setup`-only constructs like `computed`. It's already two levels of indirection (`urlReactiveAuditLogQuery -> fromQueryReactiveWithDefault -> computed`), it just makes it too easy to forget you should only be calling it in a `setup` function.

It looks like this will only ever be used once, you could just inline it into `/audit-logs` or into a component used by `/audit-logs` where it's unambiguously only ever called from `setup`

And more generally, I think this pattern of taking in functions that should only ever be called in `setup` is dicey, for the above mentioned reason that it obfuscates where things are being called.

bcspragu (approved):
LGTM! And sorry, this one slipped past, only just noticed it

frontend/lib/auditlogquery/index.ts
gbdubs: I think this isn't quite correct: I'm pretty sure (from experience around how we do editors, ex) that `computed` can be created and used outside of `setup`. One of the vue authors [seems to hold this position too](https://github.com/vuejs/core/discussions/9779#discussioncomment-7797404). Could you provide documentation for exactly what can and can't be called outside of setup? A 5 minute google left me without a clear north star.

By taking `fromQueryReactiveWithDefault` as an argument rather than invoking the composable to pull it out, we avoid invoking a composable (which I agree is a riskier game). 

My main goal was just keeping the logic for constructing + decoding URLs that point to audit log queries in one place, to avoid exporting the ~10 helper methods and constants that would be needed to replicate this interface outside of this package. I agree that we shouldn't overly obfuscate where/when core functionality is being used, but I think this approach actually does a good job of balancing between the competing goals as I understand them.

frontend/lib/auditlogquery/index.ts
gbdubs: Great catch

PR #138, opened 2024-01-12
Get the `create-report` async task in shape

### LINKED ISSUE #135: Vue warnings about setup-only functionality called outside `setup`
I noticed this in the logs: `[Vue warn]: inject() can only be used inside setup() or functional components. `, see the screenshot at the end for more details.

I think there are a few of these lurking, but this one is specifically `PortfolioListView`, and specifically `editorObjects` as used in `:value="editorObjects"`.

The core problem is that `editorObjects` is a computed which calls `portfolioEditor`, from `lib/editor/portfolio`, uses `getEditorComputedValues`, which uses all the things.

Aside from the fact this clutters our logs, it's 100% guaranteed to bite us when things mysteriously stop working. I don't know what the solution is, but we can't call `computed`/`useState`/etc after `setup` has finished running, which means they can't be invoked in `computed`s or in response to functions.

bcspragu:
This PR makes assorted tweaks to the runner's `create-report` task, like:
- Adding new `pacta.FileType`s for assorted HTML-adjacent types present in the report (see below)
  - Includes an `UNKNOWN` type in case I miss anything, we can always go back and re-label them based on file extension
- Fix some issues with paths, use the portfolio UUID as the directory prefix (to avoid collisions)
- Publish the event when we're done so the main app can record the info

Mostly unrelated changes:
- Fixes #135 by pulling `handleMissingTranslation` from the `$nuxt` object instead of `inject`
- Add a helper script for iterating quickly on the runner, `build_and_load_runner.sh`

Some resources that get created from this:

```
select * from blob WHERE blob_uri LIKE 'az://reports/%';

            id             |                                                   blob_uri                                                   | file_type |        file_name         |          created_at
---------------------------+--------------------------------------------------------------------------------------------------------------+-----------+--------------------------+-------------------------------
 blob.9c6a2be6135288dc141a | az://reports/afcc58f1-486c-437c-9fce-0a4774cd4128/index.html                                                 | html      | index.html               | 2024-01-12 22:24:07.695008+00
 blob.47508626ad7c784151d6 | az://reports/afcc58f1-486c-437c-9fce-0a4774cd4128/libs/accessible-code-block-0.0.1/empty-anchor.js           | js        | empty-anchor.js          | 2024-01-12 22:24:07.695008+00
 blob.e952045a95b60b0744fd | az://reports/afcc58f1-486c-437c-9fce-0a4774cd4128/libs/anchor-sections-1.1.0/anchor-sections-hash.css        | css       | anchor-sections-hash.css | 2024-01-12 22:24:07.695008+00
 blob.59b3c81bad6b701bbf83 | az://reports/afcc58f1-486c-437c-9fce-0a4774cd4128/libs/anchor-sections-1.1.0/anchor-sections.js              | js        | anchor-sections.js       | 2024-01-12 22:24:07.695008+00
 blob.7edd40c453bac1c53754 | az://reports/afcc58f1-486c-437c-9fce-0a4774cd4128/libs/gitbook-2.6.7/css/fontawesome/fontawesome-webfont.ttf | ttf       | fontawesome-webfont.ttf  | 2024-01-12 22:24:07.695008+00
 blob.7eda957bd02cbf3ab01f | az://reports/afcc58f1-486c-437c-9fce-0a4774cd4128/libs/gitbook-2.6.7/css/plugin-search.css                   | css       | plugin-search.css        | 2024-01-12 22:24:07.695008+00

```

Serving these files will be a follow-up PR

### DIFF
--- a/cmd/runner/README.md
+++ b/cmd/runner/README.md
@@ -15,11 +15,12 @@ bazel run //scripts:run_server -- --use_azure_runner
 
 ### Creating a new docker image to run locally
 
+When testing locally (e.g. without `--use_azure_runner`), you can build and tag a runner image locally and use that. To do that, run `bazel run //scripts:build_and_load_runner`
+
+### Cleaning up old runner containers
+
+By default, we don't auto-remove stopped containers (i.e. finished runner tasks), to give developers a chance to review the logs (e.g. with `docker logs <sha>`). To clean up all completed runs at once, run:
+
 ```bash
-# Build the runner binary
-bazel build --@io_bazel_rules_go//go/config:pure //cmd/runner:image_tarball
-# Load the new image into docker, which will output a SHA256 value
-docker load < bazel-bin/cmd/runner/image_tarball/tarball.tar
-# Tag the runner image in order for it to be picked up locally. Don't push this to the registry!
-docker tag <SHA from previous step> rmisa.azurecr.io/runner
+docker rm $(docker ps -a -q -f "status=exited" -f "ancestor=rmisa.azurecr.io/runner:latest")
 ```
--- a/cmd/runner/README.md
+++ b/cmd/runner/README.md
@@ -15,7 +15,13 @@ bazel run //scripts:run_server -- --use_azure_runner
 
 ### Creating a new docker image to run locally
 
-When testing locally (e.g. without `--use_azure_runner`), you can build and tag a runner image locally and use that. To do that, run `bazel run //scripts:build_and_load_runner`
+When developing the runner, you have two options:
+
+* **Test against local Docker** - Run the server **without** the  `--use_azure_runner`, which means async tasks will run locally, using `docker run ...`. To test local runner changes, you can build and tag a runner image locally with `bazel run //scripts:build_and_load_runner`.
+  * After running the script, the updated runner will immediately be available, no need to restart the server.
+  * This is the option you'll want to use most of the time.
+* **Test against Azure Container Apps Jobs** - Run the server **with** the  `--use_azure_runner`, which means async tasks will be run on Azure, created via the Azure API. To test changes here, you can build and tag a runner image locally with `bazel run //scripts:build_and_load_runner`, and then push it to Azure with `docker push rmisa.azurecr.io/runner:latest`
+  * You generally won't need to use this option unless you're testing something very specific about the runner's integration with Azure, as the runner code is identical whether run locally or on Azure.
 
 ### Cleaning up old runner containers
 
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -180,24 +180,68 @@ func parsePortfolioReq() (*task.ParsePortfolioRequest, error) {
 	return &task, nil
 }
 
-func (h *handler) uploadDirectory(ctx context.Context, dirPath, container string) error {
+func (h *handler) uploadDirectory(ctx context.Context, dirPath, container string) ([]*task.AnalysisArtifact, error) {
 	base := filepath.Base(dirPath)
 
-	return filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
+	var artifacts []*task.AnalysisArtifact
+	err := filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
 		if info.IsDir() {
 			return nil
 		}
 
 		// This is a file, let's upload it to the container
-		uri := blob.Join(h.blob.Scheme(), container, base, strings.TrimPrefix(path, dirPath))
+		uri := blob.Join(h.blob.Scheme(), container, base, strings.TrimPrefix(path, dirPath+"/"))
 		if err := h.uploadBlob(ctx, path, uri); err != nil {
 			return fmt.Errorf("failed to upload blob: %w", err)
 		}
+
+		fn := filepath.Base(path)
+		// Returns pacta.FileType_UNKNOWN for unrecognized extensions, which we'll serve as binary blobs.
+		ft := fileTypeFromExt(filepath.Ext(fn))
+		if ft == pacta.FileType_UNKNOWN {
+			h.logger.Error("unhandled file extension", zap.String("dir", dirPath), zap.String("file_ext", filepath.Ext(fn)))
+		}
+		artifacts = append(artifacts, &task.AnalysisArtifact{
+			BlobURI:  pacta.BlobURI(uri),
+			FileName: fn,
+			FileType: ft,
+		})
 		return nil
 	})
+	if err != nil {
+		return nil, fmt.Errorf("error while walking dir/uploading blobs: %w", err)
+	}
+	return artifacts, nil
+}
+
+func fileTypeFromExt(ext string) pacta.FileType {
+	switch ext {
+	case ".csv":
+		return pacta.FileType_CSV
+	case ".yaml":
+		return pacta.FileType_YAML
+	case ".zip":
+		return pacta.FileType_ZIP
+	case ".html":
+		return pacta.FileType_HTML
+	case ".json":
+		return pacta.FileType_JSON
+	case ".txt":
+		return pacta.FileType_TEXT
+	case ".css":
+		return pacta.FileType_CSS
+	case ".js":
+		return pacta.FileType_JS
+	case ".ttf":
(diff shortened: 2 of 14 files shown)


scripts/build_and_load_runner.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
```
gbdubs, line 2: I LOVE THIS IT WILL MAKE LIFE EASIER THANK YOU

pacta/pacta.go
```diff
@@ -215,6 +222,11 @@ var FileTypeValues = []FileType{
 	FileType_JSON,
 	FileType_HTML,
 	FileType_JSON,
+	FileType_TEXT,
+	FileType_CSS,
+	FileType_JS,
+	FileType_TTF,
+	FileType_UNKNOWN,
```
gbdubs, line 229: Nit: I don't think we want unknown in this list, since it's used to test things like convertibility to postgres, and we don't want "" to be persistable.

frontend/plugins/handle-missing-translation.ts
```diff
@@ -16,13 +16,13 @@ export default defineNuxtPlugin((nuxtApp) => {
...
   }
 
-  nuxtApp.vueApp.provide('handleMissingTranslation', handleMissingTranslation)
   const values = computed(() => {
     return missingTranslations.value
   })
   return {
     provide: {
       missingTranslations: {
+        handleMissingTranslation,
```
gbdubs, line 25: Love this, great, elegant solution.

frontend/i18n.config.ts
```diff
@@ -5,10 +5,10 @@ export default defineI18nConfig(() => ({
   missing: (locale, key, vm) => {
     // TODO(grady) figure out how to skip this if we're in production + just log.
     // Consider using process.env.NODE_ENV == 'prod', etc.
-    const fn = inject('handleMissingTranslation')
+    const fn = vm?.appContext.app.$nuxt.$missingTranslations.handleMissingTranslation
     if (fn) {
-      const callable = fn as (locale: string, key: string) => void
-      callable(locale, key)
+      // const callable = fn as (locale: string, key: string) => void
```
gbdubs, line 10: nit: remove?

frontend/i18n.config.ts
```diff
@@ -5,10 +5,10 @@ export default defineI18nConfig(() => ({
...
+      fn(locale, key)
     }
```
gbdubs, line 12: can you add a console.warn if we fall through without finding `handleMissingTranslation`?

db/sqldb/golden/schema_dump.sql
```diff
@@ -116,7 +116,12 @@ CREATE TYPE public.file_type AS ENUM (
     'yaml',
     'zip',
     'html',
-    'json'
+    'json',
+    'txt',
+    'css',
+    'js',
+    'ttf',
+    ''
```
gbdubs, line 124: Let's not add `''` as a valid enum type, we don't want to persist it.

db/sqldb/blob.go
```diff
@@ -233,11 +233,9 @@ func validateBlobForCreation(b *pacta.Blob) error {
 	if b.BlobURI == "" {
 		return fmt.Errorf("blob missing BlobURI")
 	}
-	if b.FileType == "" {
-		return fmt.Errorf("blob missing FileType")
-	}
 	if b.FileName == "" {
 		return fmt.Errorf("blob missing FileName")
 	}
+	// A blank FileType is valid, just means we don't recognize the file type yet.
```
gbdubs, line 239: I'd disagree with that - we require a file type to be set, and I think that's a pretty straightforward invariant to maintain for the course of development. Is there a reason to support this?

cmd/runner/main.go
```diff
@@ -388,13 +432,16 @@ func (h *handler) createAudit(ctx context.Context, taskID task.ID, req *task.Cre
...
 		// to be.
-		fileName := fmt.Sprintf("%d.json", i)
-		fileNames = append(fileNames, fileName)
-		destPath := filepath.Join("/", "mnt", "processed_portfolios", fileName)
+		fileNameWithExt := filepath.Base(string(blobURI))
+		if !strings.HasSuffix(fileNameWithExt, ".json") {
+			return fmt.Errorf("given blob wasn't a JSON-formatted portfolio, %q", fileNameWithExt)
+		}
+		fileNames = append(fileNames, strings.TrimSuffix(fileNameWithExt, ".json"))
+		destPath := filepath.Join("/", "mnt", "processed_portfolios", fileNameWithExt)
```
gbdubs, line 444: mnt?

cmd/runner/README.md
```diff
@@ -15,11 +15,12 @@ bazel run //scripts:run_server -- --use_azure_runner
 
 ### Creating a new docker image to run locally
 
+When testing locally (e.g. without `--use_azure_runner`), you can build and tag a runner image locally and use that. To do that, run `bazel run //scripts:build_and_load_runner`
```
gbdubs, line 18: This leaves ambiguous what happens if you're not using either of these options? Could you make that a bit clearer?

gbdubs (approved):
Woo!

cmd/runner/README.md
bcspragu: GC, made this section more explicit

cmd/runner/main.go
bcspragu: https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/mnt.html

Basically, `/mnt` is a generic place for external things being mounted in. [It's a convention Alex used in the stub repo](https://github.com/AlexAxthelm/stub_PACTA_containers_for_app/blob/190763cba67bafdf666f56eb34693da728fea21b/combined/create_report.R#L13), and it makes as much sense as any other location.

db/sqldb/blob.go
bcspragu: The context here is that the `create-report` action generates a whole directory of files, including HTML, JSON, CSS, JavaScript, fonts, and text files. I imagine the PACTA folks will want to modify these reports over time, potentially introducing new files and resources that should be served as part of the report.

If one of those new files is of a type we haven't enumerated (videos? various image formats? XML files? etc), we still want to persist the file and make a best-effort attempt to serve it. My solution is to say "well, we didn't recognize the type of the file at the time of report creation", which is `UNKNOWN`.

Happy to chat more if that doesn't make sense or you have another preferred approach.

db/sqldb/golden/schema_dump.sql
bcspragu: Discussed in the comment above, I think we do. I'm also open to calling this `unknown` or whatever you think makes sense here

frontend/i18n.config.ts
bcspragu: GC, done

frontend/i18n.config.ts
bcspragu: GC, done

pacta/pacta.go
bcspragu: ^^ discussed above

scripts/build_and_load_runner.sh
bcspragu: HAPPY TO HELP YEAH I WASNT ENJOYING DOING THIS MANUALLY EITHER

db/sqldb/blob.go
gbdubs: Yeah I guess I'm not sold there. In my mind the value of an enum is confidence you'll get back a value you expect. I think if we (at root) don't trust the data we have coming in to be of the type(s) that we expect, and we don't have a need to handle them specifically, we should simply remove the enum constraint generally, and store it as a string-based file extension.

cmd/runner/main.go
gbdubs: :+1:

db/sqldb/blob.go
bcspragu: I don't want to throw the baby out with the bathwater here. In most cases (currently all), the blob is one of a small list of types. It's just that the alternative when we encounter an unknown extension is to ignore the file entirely, which is definitely not the option we want.

And the file type enum is more generally useful than a string extension. For example, `yml` and `yaml` are both YAML files

PR #140, opened 2024-01-13
Add support for serving the static assets in a report

bcspragu:
This PR adds a new `/report/<analysis ID>` endpoint that serves a report. Importantly, this is a _server_ endpoint, our Nuxt client knows nothing about it.

That means to test it locally, you go to `localhost:8081/report/<analysis ID>`, and when deployed, it'll be `https://pacta-api.dev.rmi.siliconally.dev/report/<analysis ID>` IIRC. Since we share cookies between `*.dev.rmi.siliconally.dev`, this works fine.

The main thing missing from this PR is authz, which will be the next thing I do.

### DIFF
--- a/reportsrv/reportsrv.go
+++ b/reportsrv/reportsrv.go
@@ -41,10 +41,7 @@ type Server struct {
 }
 
 type DB interface {
-	Begin(context.Context) (db.Tx, error)
 	NoTxn(context.Context) db.Tx
-	Transactional(context.Context, func(tx db.Tx) error) error
-	RunOrContinueTransaction(db.Tx, func(tx db.Tx) error) error
 
 	AnalysisArtifactsForAnalysis(tx db.Tx, id pacta.AnalysisID) ([]*pacta.AnalysisArtifact, error)
 	Blobs(tx db.Tx, ids []pacta.BlobID) (map[pacta.BlobID]*pacta.Blob, error)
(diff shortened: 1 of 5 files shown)


reportsrv/reportsrv.go
```diff
@@ -0,0 +1,194 @@
...
+	db     DB
+	blob   Blob
+	logger *zap.Logger
+}
+
+type DB interface {
+	NoTxn(context.Context) db.Tx
+
+	AnalysisArtifactsForAnalysis(tx db.Tx, id pacta.AnalysisID) ([]*pacta.AnalysisArtifact, error)
+	Blobs(tx db.Tx, ids []pacta.BlobID) (map[pacta.BlobID]*pacta.Blob, error)
```
gbdubs, line 47: I think you're planning this for the next PR, if so great, but we'll need to generate Audit Logs here.

reportsrv/reportsrv.go
```diff
@@ -0,0 +1,194 @@
...
+		if !ok {
+			s.logger.Error("blob had invalid URI", zap.String("analysis_artifact_id", string(a.ID)), zap.String("blob_uri", uri))
+			continue
+		}
+		_, uriPath, ok := strings.Cut(path, "/")
+		if !ok {
+			s.logger.Error("path had no UUID prefix", zap.String("analysis_artifact_id", string(a.ID)), zap.String("blob_uri", uri), zap.String("blob_path", path))
+			continue
+		}
+
```
gbdubs, line 132: One piece I don't love about this approach is that if N = number of assets in a given report, we read back N rows 2*N times to serve the report. I get that this is still linear in the # of DB calls made, it just seems really duplicative, since the returns for each of these DB calls is identical. I can't think of a simple way of sidestepping this property, but I don't love it.

reportsrv/reportsrv.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		return
+	}
+
+	s.logger.Error("unknown report asset requested for analysis", zap.String("analysis_id", string(aID)), zap.String("req_path", r.URL.Path), zap.String("asset_path", subPath))
+	http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+	return
+}
+
+func fileTypeToMIME(ft pacta.FileType) string {
```
gbdubs, line 160: I guess this is one good reason to have FT saved!

gbdubs (approved):

reportsrv/reportsrv.go
bcspragu: Indeed, will do

reportsrv/reportsrv.go
bcspragu: Yeah, I also didn't like that and considered doing a dumb in-mem cache of `analysis ID -> artifacts`, but the cardinality is low (e.g. 10-20 artifacts), and the loading speed isn't a problem. Can revisit if it becomes an issue.

reportsrv/reportsrv.go
bcspragu: Indeed! As discussed, I'll update the parent PR to store unknown as `'unknown'`

Issue #148, opened 2024-01-19
Duplicate Blob URI Issue


gbdubs:
`{"error": "failed to perform operation: creating blob: creating blob row: ERROR: duplicate key value violates unique constraint \"blob_blob_uri_key\" (SQLSTATE 23505)"}`

bcspragu:
I'd imagine this happens when running an report twice on the same portfolio? IIRC, the blob URI is something like `az://reports/<portfolio UUID>/<assets...>`

gbdubs:
Oh, can we change that? We will need to be able to run reports multiple
times, specifically thinking portfolios with different properties (ESG ex),
or pacta versions.

bcspragu:
Totally, it's just this:

https://github.com/RMI-PACTA/app/blob/b743bc960caeb7be6a5b061f3ed301c46dd553e2/cmd/runner/main.go#L478-L479

The `create_report.R` script just outputs to a directory based on the input portfolio name, and we then just use that directly to create the subdir we upload the report to, but it can be set to whatever

PR #149, opened 2024-01-19
Report Buttons: View, Download + Run

gbdubs:
- Adds a shared set of buttons for viewing + dowloading a report. The download button uses `jsZip` which was the library with broadest usage that I found. Happy to supplant it for another. The zip downloads nicely, and can be opened (though the relative URLs leave the assets broken :( )
- Adds a `runAnalysis` button, with the goal being to use a standard hint in a few places to encourage folks to run audits, and to allow them to stay in the same place and visit the report once created. Works decently well, but ran into #148 , which will need independent attention
- Populates Snapshots in Analyses so we can see what they refer to.
- Populates ownerId on analysis so we can grey-out analyses that the user doesn't have access to (note this is not load-bearing authz, just used to prevent users from the frustration of an err message)
- Updates `getMaybeMe` to return back other info other than the `user`, in this case, the user's `ownerId`. This is more in line with what we do in AS, where the `me` call returns back announcements + personal tasks.
-  Fixes the `external` warning when navigating to view the report in browser
- Updates the `Metadata` sections of list views to just be an `always` standard debug until we can figure out what RMI wants out of that.
- Fixes the oapi spec's `blob_id` => `blobId`


frontend/components/analysis/AccessButtons.vue
```diff
@@ -0,0 +1,96 @@
...
+  await pactaClient.accessBlobContent({
+    items: props.analysis.artifacts.map((asset): AccessBlobContentReqItem => ({
+      blobId: asset.blob.id,
+    })),
+  }).then(async (response: AccessBlobContentResp) => {
+    const zip = new JSZip()
+    await Promise.all(response.items.map(
+      async (item): Promise<void> => {
+        const response = await fetch(item.downloadUrl)
+        const data = await response.blob()
```
bcspragu, line 48: There should be a way to stream the binary response data directly to the zip without loading it into memory first, can you see if that's an easy option?

frontend/components/analysis/AccessButtons.vue
```diff
@@ -0,0 +1,96 @@
...
+    const element = document.createElement('a')
+    element.href = URL.createObjectURL(content)
+    const fileName = `${props.analysis.name}.zip`
+    element.download = fileName
+    document.body.appendChild(element)
+    element.click()
+    document.body.removeChild(element)
+    downloadInProgress.value = false
+  })
+}
```
bcspragu, line 65: Mixing async/await and promises is never a good idea, one is just syntactic sugar for the other. I'd just commit to async/await, e.g.

```suggestion
const doDownload = async () => {
  downloadInProgress.value = true
  const response = await pactaClient.accessBlobContent({
    items: props.analysis.artifacts.map((asset): AccessBlobContentReqItem => ({
      blobId: asset.blob.id,
    })),
  })

  const zip = new JSZip()
  await Promise.all(response.items.map(
    async (item): Promise<void> => {
      const response = await fetch(item.downloadUrl)
      const data = await response.blob()
      const blob = presentOrFileBug(props.analysis.artifacts.find((artifact) => artifact.blob.id === item.blobId)).blob
      const fileName = `${blob.fileName}`
      zip.file(fileName, data)
    }),
  )
  const content = await zip.generateAsync({ type: 'blob' })

  const element = document.createElement('a')
  element.href = URL.createObjectURL(content)
  const fileName = `${props.analysis.name}.zip`
  element.download = fileName
  document.body.appendChild(element)
  element.click()
  document.body.removeChild(element)
  downloadInProgress.value = false
}
```

bcspragu (approved):

frontend/components/analysis/AccessButtons.vue
gbdubs: I played around with a few attempts here, without avail. The docs are here, nothing screams out.

https://stuk.github.io/jszip/documentation/api_jszip/file_data.html

frontend/components/analysis/AccessButtons.vue
gbdubs: SG. Done.

frontend/components/analysis/AccessButtons.vue
bcspragu: Took a look, looks like streaming is supported in Node environments, but not in the browser, c'est la vie

PR #152, opened 2024-01-20
Standardizes + Improves Related-Analysis Display

gbdubs:


frontend/components/analysis/StatusChip.vue
```diff
@@ -0,0 +1,51 @@
...
+const { t } = useI18n()
+const prefix = 'components/analysis/StatusChip'
+const tt = (s: string) => t(`${prefix}.${s}`)
+
+interface Props {
+  analysis: Analysis
+}
+const props = defineProps<Props>()
+
+const isStale = (a: Analysis): boolean => new Date().getTime() - new Date(a.createdAt).getTime() > 1000 * 60 * 10
```
bcspragu, line 13: Seems like there's a lot of overlapping logic between here and `ContextualListView`, maybe shove that into a `lib` somewhere?

bcspragu (approved):

frontend/components/analysis/StatusChip.vue
gbdubs: Great call - that was part of my goal of moving this to `StatusChip`, so I actually can fully remove it from the ContextualListView.

PR #153, opened 2024-01-21
Creates User Lookup

gbdubs:
This is one of the one-off requirements for the platform. It does lookup by user and email, and will enable merging users.


db/sqldb/migrations/0013_user_search.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
+SET pg_trgm.similarity_threshold = 0.1;
```
bcspragu, line 4: What does setting this at migration-time do? 

Reading [the docs](https://www.postgresql.org/docs/current/pgtrgm.html), I think you'd actually want to either set it on our connection pool, or on a per query basis (either in a transaction or with `SET LOCAL`)

db/sqldb/migrations/0013_user_search.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
bcspragu, line 3: This will explode on Azure if not configured there first, I've [gone ahead and set it in the portal](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-extensions), and I'll send you a Terraform PR as well.

I'm not sure if it's required locally. If not, you can remove this entirely

db/sqldb/migrations/0013_user_search.down.sql
bcspragu, line 1: If the `CREATE EXTENSION ...` ends up sticking around in the `up.sql`, we should probably `DROP EXTENSION` here for completeness

bcspragu (approved):

db/sqldb/migrations/0013_user_search.up.sql
gbdubs: OK, sounds good. Looks like it still works without it!

db/sqldb/migrations/0013_user_search.up.sql
gbdubs: Sorry I actually no longer need this - I was initially doing a fuzzy search, but I pivoted to doing an `ILIKE` search, so we no longer need similarity threshold.

db/sqldb/migrations/0013_user_search.down.sql
gbdubs: GC - removed.

db/sqldb/user.go
```diff
@@ -105,6 +106,84 @@ func (d *DB) Users(tx db.Tx, ids []pacta.UserID) (map[pacta.UserID]*pacta.User,
...
+	wheres := []string{}
+	for _, q := range qs {
+		if q.NameOrEmailLike != "" {
+			wheres = append(wheres,
+				fmt.Sprintf(
+					`name ILIKE ('%[1]s' || %[2]s || '%[1]s')
+					OR
+					canonical_email ILIKE ('%[1]s' || %[2]s || '%[1]s')`,
+					"%",
+					args.add(q.NameOrEmailLike)))
```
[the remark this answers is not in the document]
gbdubs: OK!

PR #156, opened 2024-01-21
Finishes Audit Log Interface

gbdubs:
Notable Elements:
- Gets Filtering working for all columns which was an UNDERTAKING.
- Creates a new component to make type-checking the date passing easier on `createdAt`. This also encapsulates the off-by-one error to make it so that the calendar filter will include logs from the selected dates (a reasonable default) 
- Updates the styles for `PVMultiSelect` and `PVChips` across the board to use `flex` (with Gap) rather than relying on margin bottom + left.
- Creates a `StandardFullWidth` component to put the logic for how to escape the standard-content's width constraints in a single place.
- Fixes an error where dates couldn't be encoded in the URL because of dual use of `:`
- Upgrades PrimeVue `3.32.0 => 3.32.1` to fix https://github.com/primefaces/primevue/issues/4268
- Puts the column selection into a modal
- Fixes some i18n issues
- Performs search when filters or sort changes, allowing the table interface to BE the query interface.


frontend/components/standard/FullWidth.vue
bcspragu, line 1: I maintain that all these `Standard*` things should just be `layout`s, but you've explained this to me in the past (not that I remember the logic at this moment)  so I won't harp on it

bcspragu (approved):

bcspragu (commented):
Oh wait I didn't actually review the bulk of this change, the GH UI had hidden `pages/audit-logs.vue`, looking at that now

frontend/pages/audit-logs.vue
```diff
@@ -1,7 +1,8 @@
 <script setup lang="ts">
-import { type AuditLogQuerySort, type AuditLogQueryReq, AuditLogTargetType, AuditLogQuerySortBy, type AuditLog } from '@/openapi/generated/pacta'
+import { type AuditLogQuerySort, type AuditLogQueryWhere, type AuditLogQueryReq, AuditLogAction, AuditLogTargetType, AuditLogQuerySortBy, type AuditLog } from '@/openapi/generated/pacta'
 import { urlReactiveAuditLogQuery } from '@/lib/auditlogquery'
-import { type DataTableSortMeta } from 'primevue/datatable'
+import { type DataTableSortMeta, type DataTableFilterMeta } from 'primevue/datatable'
```
bcspragu, line 4: fix the lint failure

frontend/pages/audit-logs.vue
```diff
@@ -105,6 +194,11 @@ const allColumns: Column[] = [
     field: 'action',
     header: tt('Action'),
     customBody: true,
+    filterType: FilterType.EnumBased,
+    enumValues: Object.values(AuditLogAction).map((a: AuditLogAction) => ({
+      label: `${a}`.replace('AuditLogAction', ''),
```
bcspragu, line 199: You don't need to stringify the enum, typescript is smart enough to know it's a string enum:

```suggestion
      label: a.replace('AuditLogAction', ''),
```

That said, this is fragile in the sense that it depends on some internal behavior of our TS codegen. I think it's fine in this case since its just a convenience thing, but the more robust (and tedious) approach would be to maintain the mapping explicitly

frontend/pages/audit-logs.vue
```diff
@@ -222,20 +439,31 @@ const getTargetLink = (t: AuditLogTargetType, id: string): string => {
   return '#'
 }
 
+const columnControlOverlay = useState<{ toggle: (e: Event) => void }>(`${prefix}.columnControlOverlay`)
```
bcspragu, line 442: `useState` shouldn't contain things that are unserializable, like functions, since they explode if you try to send that data over SSR

bcspragu (approved):

frontend/components/standard/FullWidth.vue
gbdubs: SG - happy to rediscuss!

PR #157, opened 2024-01-22
Tear down testpgx instances

bcspragu:
Wondering why you have 1,000 old postgres instances hanging around in Docker? Wonder no more! We were calling `os.Exit(0)` for reasons unknown to me in these scripts, and there's a very important detail in [the documentation for os.Exit](https://pkg.go.dev/os#Exit):

> Exit causes the current program to exit with the given status code. Conventionally, code zero indicates success, non-zero an error. The program terminates immediately; **deferred functions are not run.**


gbdubs (approved):
Awesome, very glad to see this fixed. Also: GTK about the behavior of os.Exit!!!

Issue #76, opened 2023-12-07
Create User-Owner at User Creation Time


gbdubs:
No need to do `GetOrCreateUserOwner` - Brandon's point in #72

PR #159, opened 2024-01-22
Recursive Deletes

gbdubs:
- Deletes analysis based on a Portfolio if the portfolio gets deleted (satisfying the constraint we've implicitly encoded through the FKC's on the `portfolio_snapshot`'s fields).
- Does the same for `portfolio_group` and `initiative` (though we can't yet run analyses based on these)
- Because this changes the signatures to return `[]pacta.BlobURI` on deletes, adds blob deletion code to the API.

NOTE: I haven't been able to validate through running locally that the deletes for PGs or Initaitives work when this is run because we don't have the ability to run portfolio group or initaitive based analyeses yet.

### DIFF
--- a/db/sqldb/initiative.go
+++ b/db/sqldb/initiative.go
@@ -114,11 +114,25 @@ func (d *DB) UpdateInitiative(tx db.Tx, id pacta.InitiativeID, mutations ...db.U
 	return nil
 }
 
-func (d *DB) DeleteInitiative(tx db.Tx, id pacta.InitiativeID) error {
+func (d *DB) DeleteInitiative(tx db.Tx, id pacta.InitiativeID) ([]pacta.BlobURI, error) {
+	buris := []pacta.BlobURI{}
 	err := d.RunOrContinueTransaction(tx, func(db.Tx) error {
-		// TODO(grady) add owner deletions here - where the initiative is the owner of the asset/blob
-		// TODO(grady) do snapshot deletions here
-		err := d.exec(tx, `DELETE FROM initiative_invitation WHERE initiative_id = $1;`, id)
+		owner, err := d.GetOrCreateOwnerForInitiative(tx, id)
+		if err != nil {
+			return fmt.Errorf("getting owner for initiative: %w", err)
+		}
+		analysisIDs, err := d.AnalysesRunOnInitiative(tx, id)
+		if err != nil {
+			return fmt.Errorf("reading initative analyses: %w", err)
+		}
+		for _, aID := range analysisIDs {
+			aBuris, err := d.DeleteAnalysis(tx, aID)
+			if err != nil {
+				return fmt.Errorf("deleting analysis: %w", err)
+			}
+			buris = append(buris, aBuris...)
+		}
+		err = d.exec(tx, `DELETE FROM initiative_invitation WHERE initiative_id = $1;`, id)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


db/sqldb/initiative.go
```diff
@@ -114,11 +114,25 @@ func (d *DB) UpdateInitiative(tx db.Tx, id pacta.InitiativeID, mutations ...db.U
...
 }
 
-func (d *DB) DeleteInitiative(tx db.Tx, id pacta.InitiativeID) error {
+func (d *DB) DeleteInitiative(tx db.Tx, id pacta.InitiativeID) ([]pacta.BlobURI, error) {
+	buris := []pacta.BlobURI{}
 	err := d.RunOrContinueTransaction(tx, func(db.Tx) error {
-		// TODO(grady) add owner deletions here - where the initiative is the owner of the asset/blob
-		// TODO(grady) do snapshot deletions here
-		err := d.exec(tx, `DELETE FROM initiative_invitation WHERE initiative_id = $1;`, id)
+		owner, err := d.GetOrCreateOwnerForInitiative(tx, id)
```
bcspragu, line 120: Seems odd that we might create an `owner` in the course of deleting something, is there not a `GetOwner...` endpoint we can use instead?

db/sqldb/owner.go
```diff
@@ -142,10 +142,13 @@ func (d *DB) DeleteOwner(tx db.Tx, oID pacta.OwnerID) ([]pacta.BlobURI, error) {
 			return fmt.Errorf("getting portfolio groups for owner: %w", err)
 		}
 		for _, pgroup := range pgroups {
-			err := d.DeletePortfolioGroup(tx, pgroup.ID)
+			pgBuris, err := d.DeletePortfolioGroup(tx, pgroup.ID)
```
bcspragu, line 145: Not sure if I called this out when this function was added, but does it make sense to delete "all the things" when deleting an owner? I.e if a user says "delete this owning user/initiative", do we want to recursively delete portfolio groups and everything beneath them? Or should we just fail and say "sorry this owner owns portfolio groups that need to be moved first"?

I ask because, while I'm glad we aren't using `CASCADE`, these functions are the equivalent of a `CASCADE` relationship and so we should make sure that the user-intention matches up with the DB intention

bcspragu (approved):

db/sqldb/initiative.go
gbdubs: Ahhh I should do the same migration I did in #76, but for initiatives. I'll do that next.

db/sqldb/initiative.go
```diff
@@ -114,11 +114,25 @@ func (d *DB) UpdateInitiative(tx db.Tx, id pacta.InitiativeID, mutations ...db.U
...
+		if err != nil {
+			return fmt.Errorf("getting owner for initiative: %w", err)
+		}
+		analysisIDs, err := d.AnalysesRunOnInitiative(tx, id)
+		if err != nil {
+			return fmt.Errorf("reading initative analyses: %w", err)
+		}
+		for _, aID := range analysisIDs {
+			aBuris, err := d.DeleteAnalysis(tx, aID)
```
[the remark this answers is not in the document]
gbdubs: Generally agree, but I think we should wait to optimize until we see concrete evidence of this not being sufficiently performant

PR #161, opened 2024-01-22
Creates Initiative Owner when creating an Initiative

gbdubs:
This mirrors the behavior on `User` creates, simplifying ownership structure.

Great suggestion Brandon!


bcspragu (approved):
Very nice!

PR #164, opened 2024-01-23
Refactors Initiative FE to support Authz + Anon Access

gbdubs:


cmd/server/main.go
```diff
@@ -372,6 +373,24 @@ func run(args []string) error {
 	return nil
 }
 
+var publicEndpoints = []*regexp.Regexp{
+	regexp.MustCompile(`^/initiative/[^/]*$`),
```
bcspragu, line 377: Ignoring my disdain for regexes for a second, I think we might want something a bit more flexible here, something like:

```go
type allowFn func(r *http.Request) bool
var publicEndpoints = []allowFn{
  allowInitiatives,
  // other stuff in the future
}

func allowInitiatives(r *http.Request) bool {
  return yourRegexp.MatchString(r.URL.path) && r.Method == http.MethodGet
}
```

The reason I suggest this is because I don't think you actually want "all paths starting with `"/initiative/"` and at most one sub-path", as that includes `POST` requests and stuff, so you'll want to check `r.Method` as well.

That said, I do also think the regex is overkill here, I'd explicitly enumerate paths instead, as that's a safer invariant as our API surface changes

cmd/server/main.go
```diff
@@ -381,6 +400,17 @@ func rateLimitMiddleware(maxReq int, windowLength time.Duration) func(http.Handl
...
+			if session.IsAllowedAnonymous(r.Context()) {
+				// Rate limit by IP address if the user is anonymous.
+				ip := r.Header.Get("X-Real-IP")
+				if ip == "" {
+					ip = r.Header.Get("X-Forwarded-For")
+				}
+				if ip == "" {
+					ip = r.RemoteAddr
+				}
+				return ip, nil
```
bcspragu, line 412: You shouldn't need to manually pull out the IP like this, [chi already does it](https://pkg.go.dev/github.com/go-chi/chi/v5@v5.0.11/middleware#RealIP). `r.RemoteAddr` should already contain what you want, as long as the chi.RealIP middleware runs first

cmd/server/pactasrv/authz.go
```diff
@@ -79,15 +79,27 @@ func notFoundOrUnauthorized[T ~string](actorInfo actorInfo, action pacta.AuditLo
 }
 
 func (s *Server) auditLogIfAuthorizedOrFail(ctx context.Context, status *authzStatus) error {
+	zapFields := []zap.Field{
```
bcspragu, line 82: You should make this a function, [like here](https://github.com/RMI/opgee-api/blob/d342c0c21c0dbf4c62025a4ee71a74094f4f37f0/cmd/apiserver/server/opgee.go#L252-L257). The reason is that Go has this very subtle footgun with [`append`](https://pkg.go.dev/builtin#append) where it may or may not modify the slice passed in.

In this case, it doesn't matter because every time you append you also immediately return, but still, some linters would call this out.

cmd/server/pactasrv/initiative.go
```diff
@@ -125,11 +129,27 @@ func (s *Server) FindInitiativeById(ctx context.Context, request api.FindInitiat
...
+		i.InitiativeUserRelationships = relationships
+	} else if actorInfo.UserID != "" {
+		iur, err := s.DB.InitiativeUserRelationship(s.DB.NoTxn(ctx), i.ID, actorInfo.UserID)
+		if err != nil {
+			return nil, oapierr.Internal("failed to load singular initiative user relationship for initiative", zap.String("initiative_id", string(i.ID)), zap.Error(err))
+		}
+		i.InitiativeUserRelationships = []*pacta.InitiativeUserRelationship{iur}
+	}
+	if !info.CanSeeInternalInfo {
+		i.InternalDescription = ""
```
bcspragu, line 151: optional: depending on how sensitive this is, I'd build up a new object, invert the condition, and only populate the sensitive stuff here. Might avoid accidentally sending down data users shouldn't have

session/session.go
```diff
@@ -13,11 +13,18 @@ import (
 )
 
 var userIDKey = struct{}{}
+var allowedAnonymousKey = struct{}{}
```
bcspragu, line 16: nit: can these be structs instead? e.g. `type userIDKey struct {}` and `type allowedAnonymousKey struct {}`, [ex](https://github.com/RMI/opgee-api/blob/d342c0c21c0dbf4c62025a4ee71a74094f4f37f0/opgee/opgee.go#L47-L51)

That's the normal shape I've seen for private context keys, using anonymous structs in this nature seems less solid.

bcspragu (approved):

cmd/server/main.go
gbdubs: Really good idea - though the reason we (probably) need regex as opposed to enumerating paths is that the second path component is the ID: `/initative/some-id-here`, and since that can be anything and be valid, we can't enumerate paths.

gbdubs: So TL;DR: I'll use the `allowFn` syntax, and also exclude non-get requests, but keep that particular regex.

cmd/server/main.go
gbdubs: Gorgeous. Thank you!

cmd/server/pactasrv/authz.go
gbdubs: Oh really smart.

cmd/server/pactasrv/initiative.go
gbdubs: I'll pass on that for now just because of the ratio of effort to value there, but ACK that would be a better way to do it, and we will do it that way for more sensitive material.

session/session.go
gbdubs: SGTM.

PR #165, opened 2024-01-23
Creates Privacy + ToS + Input Guide GDocs

gbdubs:
Uses the same approach we used on ABOUND, GDocs embedded using publishing links, so that we don't have to push new site versions for documentation updates.

### DIFF
--- a/frontend/pages/upload.vue
+++ b/frontend/pages/upload.vue
@@ -8,6 +8,7 @@ const { linkToPortfolioList } = useMyDataURLs()
 const pactaClient = usePACTA()
 const { $axios } = useNuxtApp()
 const { t } = useI18n()
+const localePath = useLocalePath()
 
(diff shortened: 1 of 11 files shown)


frontend/pages/upload.vue
```diff
@@ -296,7 +297,13 @@ const cleanUpIncompleteUploads = async () => {
     <!-- TODO(#80) Finalize this copy -->
     <p>
       This is a page where you can upload portfolios to test out the PACTA platform.
-      This Copy will need work, and will need to link to the documentation.
+    </p>
+    <p>
+      Input files are expected to be structred as CSVs with an expected set of header rows.
```
bcspragu, line 302:
```suggestion
      Input files are expected to be structured as CSVs with an expected set of header rows.
```

bcspragu (approved):

PR #166, opened 2024-01-23
FE for User Merge, Validated

gbdubs:
Wildly, it seems that User Merge works out of the box - validated that groups, portfolios and other owned assets transfer over seamlessly!

:)

### DIFF
--- a/frontend/pages/admin/merge.vue
+++ b/frontend/pages/admin/merge.vue
@@ -1,0 +1,87 @@
+<script setup lang="ts">
+const pactaClient = usePACTA()
+const { loading: { withLoading } } = useModal()
+const localePath = useLocalePath()
+
+const prefix = 'admin/merge'
+
+const fromUserId = useState<string>(`${prefix}.fromUserId`, () => '')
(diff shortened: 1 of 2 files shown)


frontend/pages/admin/merge.vue
```diff
@@ -0,0 +1,87 @@
...
+          new-tab
+        />
+      </div>
+    </div>
+    <PVButton
+      :disabled="done || !fromUserId || !toUserId || fromUserId === toUserId"
+      label="Perform Merge"
+      class="p-button-danger"
+      icon="pi pi-user-minus"
+      @click="doMerge"
```
bcspragu, line 71: optional: I feel like this should have a warning that loads metadata for source + dest users and is like:

`Are you sure you want to merge `user@abc.com` into `user2@muffins.biz`? This action is not particularly easy to undo.`

bcspragu (approved):
Congrats on the flawless execution!

frontend/pages/admin/merge.vue
gbdubs: Great call. Done

PR #171, opened 2024-01-24
Allows Anonymous Downloads of Shared Assets

### LINKED ISSUE #119: Support Public Asset Sharing

gbdubs:
Fixes #119

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -389,13 +389,11 @@ func allowPublicInitiativeLookups(r *http.Request) bool {
 	return allowPublicInitiativeLookupsRegexp.MatchString(r.URL.Path)
 }
 
-var allowPublicAnalysisDownloadsRegexp = regexp.MustCompile(`^/report/.*$`)
-
 func allowPublicAnalysisDownloads(r *http.Request) bool {
 	if r.Method != http.MethodGet {
 		return false
 	}
-	return allowPublicAnalysisDownloadsRegexp.MatchString(r.URL.Path)
+	return strings.HasPrefix(r.URL.Path, "/report/")
 }
 
 func requireJWTIfNotPublicEndpoint(next http.Handler) http.Handler {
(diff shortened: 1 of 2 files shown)


cmd/server/main.go
```diff
@@ -388,6 +389,15 @@ func allowPublicInitiativeLookups(r *http.Request) bool {
 	return allowPublicInitiativeLookupsRegexp.MatchString(r.URL.Path)
 }
 
+var allowPublicAnalysisDownloadsRegexp = regexp.MustCompile(`^/report/.*$`)
```
bcspragu, line 392: lol sorry to keep beating the anti-regexp drum here, but if we want to say "any path starting with `/report/`", that's just `strings.HasPrefix(r.URL.Path, "/report/")`

reportsrv/reportsrv.go
```diff
@@ -192,6 +192,7 @@ func (s *Server) doAuthzAndAuditLog(a *pacta.Analysis, aa *pacta.AnalysisArtifac
 	actorID, _ := session.UserIDFromContext(ctx)
 	if actorID == "" {
 		actorID = unauthenticatedUserID
+		actorOwner = &pacta.Owner{ID: unauthenticatedUserID}
```
bcspragu, line 195: I missed this before, but why set the actor to some invalid sentinel versus just leaving it blank?

bcspragu (approved):

cmd/server/main.go
gbdubs: Lol, that one is fair :laughing:

reportsrv/reportsrv.go
gbdubs: Good question: when I was thinking about the API here I realized that empty values for `actor` or `actorOwner` would be suspicious for someone on the outside to understand. Since Audit Logs are intended for external customers, saying "oh if it's unset it's just anonymous" engenders less trust than explicitly saying so in the audit log record itself. It also distinguishes between a case of missing data and a case of an anonymous user effectively.

PR #174, opened 2024-01-24
Fix error modal in user merge frontend

### LINKED ISSUE #168: Error Modal not being called in some cases
I noticed this while working on `userMerge`. Other async operations will return back the error modal if they fail, but not user merge. To reproduce, just enter bogus userIDs + try to merge them. It fails (rightly) but only console logs, and doesn't pop up the error modal.

Would you mind taking a look at this today?

bcspragu:
As noted in #168, the issue is that things thrown asynchronously (e.g. in a callback) get handled in some other context, which doesn't appear to get picked up anywhere within Nuxt/Vue.

The main way this manifests is `withLoading`, which one has to wait on to have the error thrown within the app context, though it appears in other places (like the confirm dialog, as fixed here)

To sanity check this theory, I found another un-awaited `withLoading`, on RunAnalysis. I intentionally broke the endpoint and confirmed that no error dialog appears before the changes I made in this PR.

LMK if you want to comb the codebase for these, or if you'd like me to.

Fixes #168, in two cases

### DIFF
--- a/frontend/components/analysis/RunButton.vue
+++ b/frontend/components/analysis/RunButton.vue
@@ -87,7 +87,7 @@ const runAnalysis = async () => {
       accept: () => { startRun().then(resolve).catch(reject) },
       acceptIcon: 'pi pi-check',
     })
-  });
+  })
 }
 const refreshAnalysisState = async () => {
   const aid = analysisId.value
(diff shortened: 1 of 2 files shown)


bcspragu:
Closing in favor of #177

PR #177, opened 2024-01-24
Fix error modal handling

bcspragu:
As discussed, this is an alternate approach to #174 for fixing #168. It handles errors directly in `withLoading` and `pactaClient`. The problem then is that both these functions return _something_, so you still have to throw and interrupt the flow of code (I originally tried just leaving hanging promises, but our linter informed me that's a really bad idea)

Since we have to throw, we throw a sentinel so that anything further up the chain knows not to render another dialog.

I tested it in a few different cases and confirmed it does what we expect.

### DIFF
--- a/frontend/composables/useModal.ts
+++ b/frontend/composables/useModal.ts
@@ -1,6 +1,6 @@
 import { type Ref } from 'vue'
-import { type ErrorObject } from 'serialize-error'
-import { serializeError } from 'serialize-error'
+import { type ErrorObject, serializeError } from 'serialize-error'
+import { createErrorWithRemediation, isSilent, Remediation } from '@/lib/error'
 
 export const useModal = () => {
   const prefix = 'useModal'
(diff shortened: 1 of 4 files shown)


frontend/app.vue
```diff
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { isSilent } from '~/lib/error'
```
gbdubs, line 2: nit: `@/lib/error`

frontend/app.vue
```diff
@@ -7,9 +8,11 @@ const handleError = (err: Error) => {
   if (process.client) {
     console.log(err)
   }
-  error.value = serializeError(err)
-  errorModalVisible.value = true
-  clearLoading()
+  if (!isSilent(err)) {
```
gbdubs, line 11: Solid approach!

gbdubs (approved):

frontend/app.vue
bcspragu: Lol not sure how I missed that the rest of the codebase does it that way, done

PR #181, opened 2024-02-12
Minor run_server.sh cleanups

bcspragu:
A few minor things I noticed while integrating EventGrid into OPGEE, porting back to PACTA

### DIFF
--- a/scripts/run_server.sh
+++ b/scripts/run_server.sh
@@ -4,7 +4,6 @@ set -euo pipefail
 ROOT="$BUILD_WORKSPACE_DIRECTORY"
 cd "$ROOT"
 
-# We keep it around because we'll need it at some point, but it can't be empty.
 VALID_FLAGS=(
   "with_public_endpoint"
 )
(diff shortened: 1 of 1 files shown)


scripts/run_server.sh
```diff
@@ -67,7 +65,7 @@ function create_eventgrid_subscription {
     --source-resource-id "$TOPIC_ID" \
     --endpoint-type=webhook \
     --endpoint="https://$1.${FRP_ADDR}${WEBHOOK_PATH}" \
-    --delivery-attribute-mapping "Authorization static $WEBHOOK_SHARED_SECRET true"
+    --delivery-attribute-mapping Authorization static $WEBHOOK_SHARED_SECRET true
```
gbdubs, line 68: How does this work at all? Wouldn't it need to be quoted for the secret to be interpolated/wouldn't it treat all of these as separate arguments? I trust that it works because it's in this PR, but I'm genuinely curious about why?

gbdubs (approved):

PR #182, opened 2024-02-12
Add configuration for separate parser image

bcspragu:
Basically split the "runner image" into two pieces: a parser image and a runner image. This is a pre-req for upcoming work, which will integrate the image from [1] into the pipeline

There will be an equivalent PR on the Terraform side for the updated config

[1] https://github.com/RMI-PACTA/workflow.portfolio.parsing


cmd/runner/taskrunner/taskrunner.go
```diff
@@ -20,8 +23,10 @@ type Config struct {
 	// like: /configs/{local,dev}.conf
 	ConfigPath string
 
-	// BaseImage is the runner image to execute, not specifying a tag.
-	BaseImage *task.BaseImage
+	// RunnerImage is the runner image to execute, not specifying a tag.
+	RunnerImage *task.BaseImage
```
gbdubs, line 27: Sorry I haven't paid enough attention in this space before - one of the things I thought that we wanted to record during a PACTA run was the version of the PACTA image that was used (and similarly, for some initiatives, they should use a specific tagged version for repeatable runs). That's why we have the PactaVersion object in the DB. How do you see this interacting with that? Viewing PACTA image SHAs as a binary configuration will make that less workable, and further splitting seems like it would too. What are your thoughts?

bcspragu: Yup, we're totally aligned here. You'll note that this is a `*task.BaseImage`, which specifies the registry and the name, but *not* the tag. Currently, the tag is hardcoded to `latest`, but there's a TODO (moved to the top of the file in this PR) to replace `latest` with some tag specified in the request, which would presumably come from the `PactaVersion` specified.

Does that make sense?

gbdubs: Gotcha, that makes a ton of sense. The `PactaVersion` today just has a SHA256 stored, and it sounds like that needs to be a tag instead?

bcspragu: Kinda yes, kinda no. Images can be pulled by digest, or we could tag the image with the SHA itself. The former requires minor changes here, the latter will require minor changes in our deploy process (when we get there)

bcspragu:
(approval pls)

gbdubs (approved):

PR #183, opened 2024-02-12
Split parser binary out from runner

bcspragu:
Technically, we could have used a single `runner` binary with two different base images, but that's liable to get confusing quickly, so this PR creates a new `parser` binary that is mostly just split out from the `runner`.

It also refactors a few pieces of code we've been copying around:
- `azcreds` - Logic for how to authenticate with Azure based on where we're running
- `async` - All the business logic for running async tasks, shared between `runner` + `parser`

### DIFF
--- a/async/async.go
+++ b/async/async.go
@@ -1,0 +1,372 @@
+// Package async provides the business logic for our async tasks.
+package async
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventgrid/publisher"
+	"github.com/RMI/pacta/blob"
+	"github.com/RMI/pacta/pacta"
+	"github.com/RMI/pacta/task"
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+)
+
+type Config struct {
+	Blob   Blob
+	PubSub *publisher.Client
+	Logger *zap.Logger
+}
+
+func (c *Config) validate() error {
+	if c.Blob == nil {
+		return errors.New("no blob client given")
+	}
+	if c.PubSub == nil {
+		return errors.New("no pub/sub client given")
+	}
+	if c.Logger == nil {
+		return errors.New("no logger given")
+	}
+
+	return nil
+}
+
+type Blob interface {
+	ReadBlob(ctx context.Context, uri string) (io.ReadCloser, error)
+	WriteBlob(ctx context.Context, uri string, r io.Reader) error
+	Scheme() blob.Scheme
+}
+
+type Handler struct {
+	blob   Blob
+	pubsub *publisher.Client
+	logger *zap.Logger
+}
+
+func New(cfg *Config) (*Handler, error) {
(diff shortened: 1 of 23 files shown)


async/async.go
```diff
@@ -0,0 +1,372 @@
...
+		if idx == -1 {
+			continue
+		}
+		paths = append(paths, strings.TrimSpace(line[idx+17:]))
+	}
+
+	// NOTE: This code could benefit from some concurrency, but I'm opting not to prematurely optimize.
+	var out []*task.ParsePortfolioResponseItem
+	for _, p := range paths {
+		lineCount, err := countCSVLines(p)
```
gbdubs, line 114: Let's remove this - in the discussion with RMI on the doc I sent over last week we agreed that any metrics displayed to the user should come from them.

async/async.go
```diff
@@ -0,0 +1,372 @@
...
+	if _, err := h.pubsub.PublishEvents(ctx, events, nil); err != nil {
+		return fmt.Errorf("failed to publish event: %w", err)
+	}
+
+	h.logger.Info("parsed portfolio", zap.String("task_id", string(taskID)))
+
+	return nil
+}
+
+// TODO(grady): Move this line counting into the image to prevent having our code do any read of the actual underlying data.
```
gbdubs, line 162: Great minds - yeah lets' just remove this.

async/async.go
```diff
@@ -0,0 +1,372 @@
...
+	case ".txt":
+		return pacta.FileType_TEXT
+	case ".css":
+		return pacta.FileType_CSS
+	case ".js":
+		return pacta.FileType_JS
+	case ".ttf":
+		return pacta.FileType_TTF
+	default:
+		return pacta.FileType_UNKNOWN
```
gbdubs, line 350: Can we add a log here for missing cases?

async/req.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, errors.New("no CREATE_AUDIT_REQUEST was given")
+	}
+	var task task.CreateAuditRequest
+	if err := json.NewDecoder(strings.NewReader(car)).Decode(&task); err != nil {
+		return nil, fmt.Errorf("failed to load CreateAuditRequest: %w", err)
+	}
+	return &task, nil
+}
+
+func CreateReportReq() (*task.CreateReportRequest, error) {
```
gbdubs, line 37: nit: Consider renaming these from `CreateX` to something more like `GetXFromEnvVars` - it's just odd to see a `Create` method without any caller-discernable input.

azure/azcreds/azcreds.go
```diff
@@ -0,0 +1,71 @@
+// Package azcreds provides helpers for getting environment-appropriate credentials.
```
gbdubs, line 1: This file is super clear and really well documented. Well done!

gbdubs (approved):

async/async.go
bcspragu: Excellent news for you (which you've likely seen at this point): The updated parser code does surface this info, and this is indeed removed there.

async/async.go
bcspragu: See above

(also, not great minds, just one great mind. this is literally your comment, just moved)

async/req.go
bcspragu: GC, done

async/async.go
bcspragu: Indeed we do! The method is kept simple for separation of concern purposes, but we do indeed log an error if the extension is `UNKNOWN`

PR #186, opened 2024-02-14
Finish integrating new parser into pipeline

bcspragu:
This PR finishes integrating the new parser logic from [1] into our pipeline.

It parses the `processed_portfolios.json` file from the output directory (in this case `/home/portfolio-parser/output`) and uses that to both correlate input + output files as well as upload the output CSV files. Since the R code now includes a row count, we no longer need to parse the files manually.

This all mostly works as expected. A few sharp edges (relying on UUIDs from the R code) are noted in the PR, and there's metadata produced by the new code (both at the input file level and the output file level) that we aren't currently recording anywhere.

Adjacent changes:
- In creating the `parser`, I also duplicated the `taskrunner` package. That has been hoisted to the top level and de-duped
- Assorted refactorings and renamings to make sure the `pactaparser` image gets invoked correctly

[1] https://github.com/RMI-PACTA/workflow.portfolio.parsing

### DIFF
--- a/async/async.go
+++ b/async/async.go
@@ -23,6 +23,7 @@ import (
 	"github.com/RMI/pacta/task"
 	"github.com/google/uuid"
 	"go.uber.org/zap"
+	"go.uber.org/zap/exp/zapfield"
 )
 
 type Config struct {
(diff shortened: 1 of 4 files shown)


async/async.go
```diff
@@ -70,69 +71,101 @@ func New(cfg *Config) (*Handler, error) {
...
-		if err := h.uploadBlob(ctx, p, string(blobURI)); err != nil {
-			return fmt.Errorf("failed to copy parsed portfolio from %q to %q: %w", p, blobURI, err)
-		}
-		extension := filepath.Ext(fileName)
-		fileType, err := pacta.ParseFileType(extension)
-		if err != nil {
-			return fmt.Errorf("failed to parse file type from file name %q: %w", fileName, err)
+	for _, sf := range sourceFiles {
+
+		// TODO: There's lots of metadata associated with the input files (e.g.
```
gbdubs, line 135: Can you create a TODO to track this? I feel like capturing this data and storing it seems like it is more likely than not to save us time and energy down the road if/when something goes wrong. Storing it as a blob (using our existing infrastructure) and then either attaching that blob to the `incomplete_upload` or creating a new table for it is probably the right call.

async/async.go
```diff
@@ -70,69 +71,101 @@ func New(cfg *Config) (*Handler, error) {
...
+		// sf.Errors, sf.GroupCols, etc), we should likely store that info somewhere.
+
+		for _, p := range sf.Portfolios {
+			outPath := filepath.Join(outputDir, p.OutputFilename)
+
+			// XXX: One risk here is that we're depending on the R code to generate truly
```
gbdubs, line 141: Why rely on that at all? The info we actually need for semantics is in the portfolio metadata. Let's create UUIDs or similar here?

async/async.go
```diff
@@ -70,69 +71,101 @@ func New(cfg *Config) (*Handler, error) {
...
+			// random UUIDs, we likely want some sort of namespacing here.
+			blobURI := pacta.BlobURI(blob.Join(h.blob.Scheme(), destPortfolioContainer, p.OutputFilename))
+
+			if err := h.uploadBlob(ctx, outPath, string(blobURI)); err != nil {
+				return fmt.Errorf("failed to copy parsed portfolio from %q to %q: %w", p, blobURI, err)
+			}
+			extension := filepath.Ext(p.OutputFilename)
+			fileType, err := pacta.ParseFileType(extension)
```
gbdubs, line 149: I think this can be made more robust, no? We expect this to be a CSV and probably want it to error if it is not.

async/async.go
```diff
@@ -70,69 +71,101 @@ func New(cfg *Config) (*Handler, error) {
...
+			if err != nil {
+				return fmt.Errorf("failed to parse file type from file name %q: %w", p.OutputFilename, err)
+			}
+
+			sourceURI, ok := localCSVToBlob[sf.InputFilename]
```
gbdubs, line 154: Optional: consider adding another check that no additional output files are present that aren't mentioned in the manifest. That kind of defensive programming is probably warranted over a trusted but blackbox interface like this one.

(after the loop completes, obviously. A map/set should be sufficient)

async/async.go
```diff
@@ -70,69 +71,101 @@ func New(cfg *Config) (*Handler, error) {
...
-	for sc.Scan() {
-		line := sc.Text()
-		idx := strings.Index(line, "writing to file: " /* 17 chars */)
-		if idx == -1 {
-			continue
-		}
-		paths = append(paths, strings.TrimSpace(line[idx+17:]))
+	var sourceFiles []parsed.SourceFile
+	if err := json.NewDecoder(omf).Decode(&sourceFiles); err != nil {
+		return fmt.Errorf("failed to decode processed_portfolios.json as JSON: %w", err)
```
gbdubs, line 128: If this step fails, we probably want a more complete accounting of why. Would logging the full manifest as a string if this fails be inappropriate? Maybe upload it to a cloud bucket?

async/parsed/parsed.go
```diff
@@ -0,0 +1,35 @@
+// Package parsed just holds the domain types for dealing with the output of the
```
gbdubs, line 1: Perhaps link to the repo that this relies upon? It's not obvious from this comment that this describes the contours of an external dependency.

azure/azevents/azevents.go
```diff
@@ -327,10 +327,13 @@ func (s *Server) handleParsedPortfolio(id string, resp *task.ParsePortfolioRespo
 			if err != nil {
 				return fmt.Errorf("creating blob %d: %w", i, err)
 			}
+
+			// TODO: There's other metadata in output.Portfolio, like `InvestorName`, that
```
gbdubs, line 331: Can you add a bug for this? I don't want to lose track of it.

gbdubs (approved):
:+1: Psyched to see the complete coupling come into focus!

async/async.go
bcspragu: This is a great call. The R code actually already does log the output `processed_porfolios.json` file to stdout (or stderr), so this will be covered by #185

async/async.go
bcspragu: Done, see #187

async/async.go
bcspragu: It was mostly just because there's already a bunch of UUIDs flying around and being mapped to things and I didn't want to introduce more cognitive load with another opaque UUID -> UUID mapping, but agreed that we shouldn't rely on this behavior since we don't have to.

Done, and also simpler than I thought, we don't need to maintain the mapping as we just store the blob URI as is.

async/async.go
bcspragu: I considered this and then got lazy, this is done now.

async/parsed/parsed.go
bcspragu: GC, done

azure/azevents/azevents.go
bcspragu: Done

bcspragu (commented):
Good calls across the board. You've once again caught all the small validations and other things I cheaped out on, and the code is better for it!

